PR #236, opened 2015-01-10
MRG: Add initial terms and examples for NIDM Experiment

nicholsn:
This PR implements an initial set of examples and terminology for the NIDM Experiment component.

A few noteworthy comments:

1. Metadata providing the actual vocabulary with a title and description
2. PROV and Ontology Metadata ontologies are embedded rather than imported, following the OBI convention. Note: this could be done instead in a 'nidm-core' vocabulary
3. Ontology Metadata is included for terms indicating their curation status 
4. Object restrictions are used on classes, rather than domain and range of properties for more correct semantics
5. http://purl.org/nidash/nidm/experiment is used as a namespace to actually dereference to the location of the terms file on github (note: this can be changed and is a temporary work around until INCF implements a redirection or content negotiation service)

Note: the examples and terminology aren't 100% in sync and some work is still to be done on the examples. My hope is to raise these as issues and work on them after the initial merge.


khelm:
Hi Nolan,

   Can you discuss what you see going into each of the following files:
    D nidm/nidm-experiment/acquisition.ttl (50)
    R nidm/nidm-experiment/openfmri/nidm-exp-acquisition.ttl (0)
    R nidm/nidm-experiment/openfmri/nidm-exp-project.ttl (0)
    R nidm/nidm-experiment/openfmri/nidm-exp-study-mr.ttl (0)
    R nidm/nidm-experiment/openfmri/nidm-exp-study-ques.ttl (0)

For example, what I have been working on lately (and what Jess and I started at the last meeting) is imaging modality terms and working a bit up the tree just to see where we would have to make some decisions about OBI/DOLCE.  The DICOM stuff will obviously be attached to those terms.  Do you see all of that as exp-acquisition?

nicholsn:
> Can you discuss what you see going into each of the following files:

@khelm, sure happy to! I am going to reorder them to address top down. Also note that in nidm-experiment.owl I have included preliminary definitions.

> R nidm/nidm-experiment/openfmri/nidm-exp-project.ttl (0)

In this file I would like to capture the type of information that you would find on a project's homepage. For example, I would like to be able to generate a web page that provides the title, description, affiliated investigators, data license, etc., as well as links to the specific subject or studies that were conducted for this project.

> R nidm/nidm-experiment/openfmri/nidm-exp-study-mr.ttl (0)
> R nidm/nidm-experiment/openfmri/nidm-exp-study-ques.ttl (0)

For Study objects, I want to capture a collection of observations that took place during a given chuck/epoch of time. For questionnaires or neuropsych, this may be the initial collection of demographics or a battery of neuropsych tests that are administered together, where each instrument is a separate acquisition.

For an MRI study, this would be an activity over a period of time from when a subject entered the scanner to when they came out, so it would include all the scans (including other observations like physio, eeg, etc.) which would be listed as acquisitions.

At the study level the information modeled does not change from one acquisition to the next - for example, at the study level for MRI you may include DICOM metadata for the date, participant identifier, scanner manufacturer, and a protocol for what the expected scan series are. 

> D nidm/nidm-experiment/acquisition.ttl (50)

This one was deleted - it was just a generic representation.

> R nidm/nidm-experiment/openfmri/nidm-exp-acquisition.ttl (0)

At the acquisition level, and this example could/should be split up by acquisition type (i.e., neuropsych, MR, etc.), a representation of the actual observation is recorded. This includes an item level representation of neuropsych data (basically a row in a table), and for MRI it would include a download link and a representation of the DICOM metadata that is specific to the given series. Essentially I want to capture a what is extracted by DCMStack: https://dcmstack.readthedocs.org/en/latest/DcmMeta_Extension.html#example

Not so sure about DOLCE (last release was in 2006, right?), but yes - the terms you are curating would provide semantic annotations for DICOM metadata at the acquisition level.

I've also been looking at the Semantic Science Ontology, which is comparable to OBI: https://code.google.com/p/semanticscience/wiki/SIO

Note: I'm not tied super tied to this representation, but it basically reuses the concepts from XCEDE and will let me map from XNAT/REDCap and query w/o too much ontological overhead.

khelm:
Thanks, @nicholsn.  Are you looking at project/study/acquisition as categories of metadata as well as directory hierarchies or just the former?  

The reason I'm asking is that I'm currently setting up the storage hierarchy for the multi-modal project I'm working on and so want to follow the XCEDE-like hierarchy you outlined above.  I'm in a bit of a different situation as I'm using a document store, couchdb, so I don't have a schema, but I could group the metadata that I'm extracting into project/study/acquisition objects.

However, I also want to use the metadata to determine where the large data files are stored on the file system.  For the smaller files (behavioral, spreadsheets, etc.), I am creating JSON metadata documents to be stored in the database and saving the file as an attachment.  So I could use project/study/acquisition hierarchy for the filesystem, but would have to modify it to accommodate the different modalities (so one extra layer).

khelm:
And as to which foundational ontology to use: I don't have a huge preference.  As I said at the meeting, I chose DOLCE because I was interested in seeing what NeuroOntoLog had already done and building off of that where possible.  However, we built the graph backwards in some cases only to see where that led us in DOLCE and the other foundational ontologies they used.  It's not set in stone on our side either.  The Semantic Science Ontology seems very similar to what we used in the parts that are most important to us and the relation hierarchies as well.  However, Jess and I borrowed liberally from the OBI relation hierarchies and didn't restrict ourselves to one source.  (IIRC...right, Jess?)

cmaumet:
I think it would be good to merge this and continue specific discussions in separate threads as suggested by @nicholsn.

@nicholsn: once you have accepted the small fix included in nicholsn/nidm#9, I would be happy to merge this one.

nicholsn:
@cmaumet, just merged - thanks for catching that!

cmaumet:
I will now merge this PR. I created a couple of separate issues to discuss specific points (i.e. #237, #238 and #240), please feel free to create more if you think I missed some topics...

PR #242, opened 2015-01-24
Merge: External ontology terms imports

cmaumet:
In this pull request:
 - `nidm-results.owl` and `nidm-experiment.owl` are updated to import their common set of terms coming from IAO from `iao_import.ttl`. 
 - `iao_import.ttl` was created using http://ontofox.hegroup.org/ (implementing the MIREOT approach) based on `iao_import.txt`. 
 - Moreover the full PROV ontology is imported in both `nidm-results.owl` and `nidm-experiment.owl` (so that we do not need to manually copy-paste PROV terms).

(This was inspired from what is done in STATO).

@nicholsn: what do you think about this?


nidm/nidm-results/terms/catalog-v001.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<catalog prefer="public" xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog">
+    <uri id="Imports Wizard Entry" name="http://www.w3.org/ns/prov-o-20130430" uri="http://www.w3.org/ns/prov-o-20130430"/>
+    <uri id="Imports Wizard Entry" name="http://purl.obolibrary.org/nidm/iao_import.owl" uri="../../imports/iao_import.ttl"/>
```
nicholsn: should http://purl.obolibrary.org/nidm/iao_import.owl be http://purl.org/nidash/nidm/iao_import.owl?

nicholsn:
@cmaumet, this looks really nice! My only question is about using owl:imports rather than simply appending the external ontologies/terms that you are capturing with mireot directly in nidm-experiment.owl.

In OBI, I believe the release/master version simply includes all the external terms in the ontology rather than using imports. To see what I mean, open OBI in protege and look at the imports - same thing for STATO.

Other than that, I think this is great!

nidm/nidm-results/terms/catalog-v001.xml
cmaumet: Thanks for catching that! Should be fixed by 1476aa9.

cmaumet:
Hi @nicholsn: you are right STATO is actually releasing files including both the STATO and the imported terms. 

I think it still makes sense to have a separate import file when working on the development version. We then get a clearer view of our custom terms and we can re-use the same import file across components. Would that work for you?

cmaumet:
Sorry, I forgot to mention: we could then include the imported terms in the ontology each time we do a release (stored in [releases folder](https://github.com/incf-nidash/nidm/tree/master/nidm/nidm-results/terms/releases)).

nicholsn:
sounds reasonable to me

cmaumet:
I think I will merge this tomorrow morning before we start converting the NIDM-experiment model into RDF...

nicholsn:
I can merge it - this will be good to work on

PR #244, opened 2015-01-27
MR datasets ontology by @khelm and @drjturner

cmaumet:
First version made at the Washington Task force meeting.


nicholsn:
Looks good

PR #246, opened 2015-02-18
NIDM-Results specification: re-organisation in 4 sub-sections

cmaumet:
This pull request introduces the following updates to the `NIDM-Results` specification:
- Divide the specification in 4 subsections: 
  1. Description of generic entities used throughout the specification (e.g. `Map`)
  2. Description of entities used around the first activity `nidm:ModelParametersEstimation`
  3. Description of entities used around the second activity `nidm:ContrastEstimation`
  4. Description of entities used around the third activity `nidm:Inference`
- When a class is used as range by an Object Property (e.g. `nidm:EstimationMethod` is the range for `nidm: withEstimationMethod `, include the description of this class just after the first use of the corresponding ObjectProperty.

PR #247, opened 2015-02-18
Re-use of Dublin Core terms

cmaumet:
This is a proposal to re-use Dublin Core terms (instead of creating custom `nidm` terms) for the following concepts (not specific to neuroimaging):
- [dctype:Image](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#dcmitype-Image) (defined as "A visual representation other than text") instead of `nidm:Image` (defined as "Class used to represent png, gif...`").
- [dct:identifier](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#terms-identifier) (defined as "An unambiguous reference to the resource within a given context. Comment:	Recommended best practice is to identify the resource by means of a string conforming to a formal identification system.") instead of `nidm:filename` (same as OBIws_0000001).
-  [dct:abstract](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#terms-abstract) (defined as "A summary of the resource.") instead of `nidm:visualisation` (defined as "Graphical representation of an entity").

We are already using [dct:format](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#terms-format) to describe file formats so I think it would make sense to re-use other Dublin Core terms as proposed in this PR. Please let me know what you think!


nicholst:
This looks good; the only one that gives me pause is `dct:abstract` for `nidm:visualisation` ("Graphical representation of an entity"), but then I can't say I really like the term  `nidm:visualisation` (we visualise lots of things).

So... won't say change anything, but wondering how others feel about it.

cmaumet:
(Just rebased on #246)

satra:
@cmaumet - i'm a little worried about `dct:identifier` - it doesn't seem to clarify what `context` is. we all know how unambiguous `cope1.nii.gz` is!

nicholsn:
I don't think of a filename as an identifier, although I can imagine using
for the dataset numbers (e.g., ds000001)... Not clarifying context is
possibly an issue - in this example context could potentially be a single
NIDM document.

I've also considered `dct:title` and `dct:description` as useful general
purpose attributes, esp. if you think of rendering a given object as a
webpage.

cmaumet:
@satra, @nicholsn: agreed `dct:identifier` is probably not very clear for a file name... I would suggest we go back to OBO:"filename" (discussed in #169). Another alternative would be to use [dct:title]( http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#title) (which is nice because we could use [dct:alternative](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#alternative) for software-specific filename e.g. `dct:title = "Contrast.nii.gz"` and `dct:alternative = "cope1.nii.gz"`).

nicholsn:
+1

satra:
i like `dct:title` as well.

khelm:
I vote for using the OBO: "filename" rather than dct:"title", since it is the former.  Granted "title" is a name of something, but I've never seen it used in reference to a file.

+1 for dctype:"Image"

I agree that @nicholst that dct:abstract is not a direct replacement for nidm:"visualization". How is "visualization" used?

jbpoline:
dctype:"Image" sounds good - but are we sure this does not imply some restriction - basically used only when talking about a 2D png / gif .. ? 
+1 for OBO:"filename"

cmaumet:
@nicholst, @khelm: [dct:description](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#description) (parent of `dct:abstract`) defined by "An account of the resource. Comment:	Description may include but is not limited to: an abstract, a table of contents, a graphical representation, or a free-text account of the resource." might be a better replacement for `nidm:visualisation`. How would you feel about it? 

`nidm:visualisation` is currently used in two places: 1. to link the `DesignMatrix` entity to a png image of the design matrix; 2. to link the `ExcursionSet` entity to a png image of the excursion set.

@nicholsn, @satra, @khelm, @jbpoline: I think that we could find examples in which Dublin core `dct:identifier` (e.g. [here](http://www.olemiss.edu/depts/general_library/dac/files/user_guide.html#title)) and `dct:title` are used to store file names. But I agree with @jbpoline and @khelm that the meaning might be mis-leading... I would say +1 for `obo:filename`, though if we find a more generic replacement later on I would prefer to go for it.

@jbpoline: `nidm:Image` is currently only used to represent a 2D png/gif (e.g. the image of the design matrix) so I would say that the dublin core `dct:Image` is a good fit. What do you think?

dbkeator:
+1 for OBO:filename, +1 for not using dct:abstract in place of nidm:visualisation (unless there are additional compelling reasons), +1 for dct:title, and +1 for dctype:image but I agree with @jbpoline in that this might be restricted to 2D images so should check.

cmaumet:
We discussed this issue on NIDM call on 23rd of Feb, 2015:
- we agreed on `dctype:Image` in replacement to `nidm:Image`

In the discussion on `nidm:visualisation`:
- @nicholsn suggested to look at: http://schema.org/ImageObject.

This will be discussed again next week. Everyone to look at schema.org and Dublin core.

cmaumet:
You will find below suggested replacements for `nidm:Image`, `nidm:visualisation`, `nidm:filename` found in [Dublin core](http://dublincore.org/documents/dcmi-terms/), [schema.org](http://schema.org/) and  [NEPOMUK File Ontology](NEPOMUK File Ontology) (suggested by email by @jbpoline):

###### 1. [Dublin core](http://dublincore.org/documents/dcmi-terms/)
- [nidm:Image](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-image) => [dctype:Image](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#dcmitype-Image): "A visual representation other than text."
- [nidm:visualisation](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-visualisation) => [dct:description](http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=terms#description): "An account of the resource. Comment: Description may include but is not limited to: an abstract, a table of contents, a graphical representation, or a free-text account of the resource."
- [nidm:filename](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-filename) => [obo:filename](http://purl.obolibrary.org/obo/OBIws_0000001): " A textual entity that refers to a digital document derived from a computer system."

###### 2. [schema.org](http://schema.org/)
- [nidm:Image](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-image) => [sch:ImageObject](http://schema.org/ImageObject): "An image file."
- [nidm:visualisation](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-visualisation) => [sch:image](http://schema.org/image): "An image of the item. This can be a URL or a fully described ImageObject."
- [nidm:filename](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-filename) => [sch:name](http://schema.org/name): "The name of the item."
 
###### 3. [NEPOMUK File Ontology](NEPOMUK File Ontology)
- [nidm:Image](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-image) => [nfo:Image](http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/#Image): "A file containing an image."
- [nidm:visualisation](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-visualisation) => [nfo:depiction](http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/#depiction): "Relates an information element to an image which depicts said element."
- [nidm:filename](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-filename) => [nfo:fileName](http://www.semanticdesktop.org/ontologies/2007/03/22/nfo/#fileName): "Name of the file, together with the extension."

+1 for NFO that provides the definitions closest to our needs (in particular I like the explicit mention of **together with the extension** in nfo:filename).

cmaumet:
We discussed this on NIDM call on March 2nd:
 - everyone to vote on the issue: to select between Dublin core/schema.org/NFO.

drjturner:
I'd say use Dublin Core for broader re-use, and pull from NFO as synonyms/alternative IDs where possible.

tiborauer:
+1 for NFO

nicholst:
+1 for Dublin core.  I'm not really bothered about this, but as the widely used Dublin core doesn't seem hugely off, I think we might as well stick to it.

jbpoline:
I like Jess idea - or could it be done the other way : use NFO and link to DCT ?

dbkeator:
+1 DCT.  I think the broader use is better.  I wouldn't use dct:abstract nidm:visualization though.

nicholsn:
+1 for Dublin Core

sent via mobile, please excuse brevity and typos

khelm:
+1 for dctype:Image - good definition, but I could live with the nfo:Image too.
+1 for nfo:depiction and nfo:fileName.  I like the "with extension" bit too, Camille. 

And:
I vote for using the NFO terms that we like and then, if it goes away in the future, we make them our own terms.  I don't see the downside to that, nor any real advantage to dct, just because dct is more well known.

cmaumet:
Thank you all for your feedback! Following the vote, it seems that Dublin Core is slightly more popular. Re-use of DC terms is now implemented.

Regarding `nidm:filename` (for which there was no good replacement in Dublin Core), looking back at `obo:digital object filename`, it looks like this term is a class rather than a data property... I have therefore re-used `nfo:fileName` as proposed by @khelm. 

If there are no further comments I think this is good to merge.

PR #248, opened 2015-02-19
Hemodynamic Response Function (HRF)

cmaumet:
This is a proposal written up with @nicholst regarding the modelling of the hemodynamic response function in first-level design matrices. We identified the need to differentiate between 3 scenarios:
 1. HRF Model using a **Single HRF kernel** (e.g. "SPM canonical HRF", "FSL Gamma HRF")
 2. HRF Model using a **Basis, one being HRF, other derived** (e.g. "SPM informed basis set")
 3. HRF Model using a **Basis, no one element corresponding to actual ideal HRF** (F-tests only) (e.g. "Finite Impulse Response Basis").

###### 1. Single HRF kernel
- Add an attribute `nidm:hasHemodynamicResponseFunction` to the `DesignMatrix`.

| Example  | Possible values for `hasHemodynamicResponseFunction`|
| ------------- | ------------- |
| dm1|ca4a9e70-b856-11e4-9072-1c6b47641236|

###### 2. Basis, one being HRF, other derived
- Add an attribute `nidm:hasHemodynamicResponseFunction` to the `DesignMatrix`.
- Add an attribute `nidm:hasHemodynamicResponseFunctionDerivative` to the `DesignMatrix` for each derivative.

| Example  | Possible values for `hasHemodynamicResponseFunctionDerivative`|
| ------------- | ------------- |
| dm2| bdf0063c-b857-11e4-93a8-873737485563 |

###### 3. Basis, no one element corresponding to actual ideal HRF
- Add an attribute `nidm:hasHemodynamicResponseBasis` to the `DesignMatrix`.

| Example  | Possible values for `hasHemodynamicResponseBasis`|
| ------------- | ------------- |
| dm3|f4625abc-b857-11e4-8e30-98beb42cbd48 |

Definitions and refining of the terms naming will be discussed in a separate issue once we have agreed on the overall modelling.

(The need to model the cross-software differences in HRF was initially discussed with part of the members of @afniHQ).


tiborauer:
Do we also model the different derivatives (spatial, temporal)?

nicholst:
In case it isn't clear:  At this stage we're simply trying to make a note of this as a global property of the design, inspired by the best practice methods reporting suggestions in [Poldrack 2008](http://www.ncbi.nlm.nih.gov/pubmed/18191585).  We are *not* doing anything close to real provenance of the design (e.g. not worrying about onset times, etc).

@tiborauer "informed basis" is what the SPM Manual calls the HRF+TemporalDeriv+DispersionDeriv.  So, yes, it's in there; see also "HemodynamicResponseFunctionDerivative".

We could perhaps have a more informative description... '1st order Taylor series expansion of HRF with respect to delay', '1st order Taylor series expansion of HRF with respect to dispersion', as that's *why* we're taking the derivatives.

@cmaumet: Why does the pictorial "TemporalDerivative" have two sub-elements?  Are those the SPM and the FSL versions of the HRF Temporal Derivative?

cmaumet:
@nicholst: yes, duplicates correspond to the temporal derivatives in `nidm`, `spm` and `fsl` namespaces (the figures are now updated to make this more explicit).

jbpoline:
I like that. The only question I have is whether we could not have something simpler / more general:

nidm:hasHRF            "True"|"False"
nidm:HRFbasis         "spm:GammaDifference"
(a basis of one function is fine ;)
or
nidm:HRFbasis         "spm:GammaDifference"
nidm:HRFbasis         "spm:HRFTemporalDerivative"
nidm:HRFbasis         "spm:HRFDispersionDerivative"
or 
nidm:HRFbasis         "spm:FourrierBasis"
etc...

Or, even simpler, if hasHRF "True" is redundant with the presence of a "hasHRF":
nidm:hasHRF         "spm:GammaDifference"
nidm:hasHRF         "spm:HRFTemporalDerivative"
or 
nidm:hasHRF         "spm:FourrierBasis"
etc...

nicholst:
@jbpoline, I think I get it, but then are you saying we'd indicate the multiple elements of a HRF basis with multiple nidm:HRFBasis entries?  Is that allowed? 

-Tom

jbpoline:
Well, the original proposal had:

hasHRFDerivative        "TemporalDerivative"
hasHRFDerivative        "DispersionDerivative"

so I assumed we could go this way.

cmaumet:
@jbpoline: this is also fine with me. The only down-side is that it makes the query to extract those studies where one of the basis represented an idealised HRF more complex.

dbkeator:
Question, when one issues a SPARQL query against an attriubute "hasHRFDerivative" or "hasHRF" where there are multiple similarly named attributes, will the query return all of them or just the first one it encounters for each entity?

satra:
@dbkeator - depends on the query. in general, sparql will only return things you ask for. 

you can use optional or filter using regex.

cmaumet:
We discussed this on NIDASH call on 23rd of February:
- we agreed on @jbpoline's suggestion to have a single property `nidm:hasHRFBasis`.

satra:
perhaps just: `nidm:hasBasis` this would cover crazy things like FLOBS and FIR and other such?

jbpoline:
I think I would prefer hasHRFbasis for clarity of the term - FIR does fall in that category - but no strong feeling on this.

nicholst:
I think I concur that HRFBasis is better. FIR is also a HRF basis set, as is FLOBS (they are usually convoluted with s delta function & so it may not seem so.)

nicholst:
@cmaumet  - We might as well add AFNI to this:  AFNI's default HRF is a single gamma (-GAM option to [3ddeconvolve](http://afni.nimh.nih.gov/pub/dist/doc/program_help/3dDeconvolve.html), or implied via [afni_proc.py](http://afni.nimh.nih.gov/pub/dist/doc/program_help/afni_proc.py.html), though it can also replicate SPM's HRF options with -SPMG1, -SPMG2, -SPMG3 (for canonical only, canonical+Temp, canonical+Temp+Disp).   (@afni-rickr - please say if I've mangled anything here.)

cmaumet:
In 6c1381b:
- As agreed on NIDM call: keep only a single object property `nidm:hasHRFBasis` used in all scenarios (1, 2 and 3).
- As suggested by @nicholst: add a term to describe afni HRF: `afni:GammaHRF` (I was thinking we could re-use the spm terms for the `-SPMG*` options. We can discuss this further with @afni-rickr).
- The class hierarchy is updated accordingly:
image
(Some of the classes will become individuals when we discuss the definitions). 

Unless there are further comments, I think this is good to merge? I will then open a separate issue to focus on the definitions.

cmaumet:
+1 to use `nidm:hasHRFBasis` to have an explicit mention of "HRF".

cmaumet:
In 6c1381b:
- As agreed on NIDM call: keep only a single object property `nidm:hasHRFBasis` used in all scenarios (1, 2 and 3).
- As suggested by @nicholst: add a term to describe afni HRF: `afni:GammaHRF` (I was thinking we could re-use the spm terms for the `-SPMG*` options. We can discuss this further with @afni-rickr).
- The class hierarchy is updated accordingly:
image
(Some of the classes will become individuals when we discuss the definitions). 

Unless there are further comments, I think this is good to merge? I will then open a separate issue to focus on the definitions.

nicholst:
+1 to merge

-Tom

satra:
i don't have a strong feeling here, just hoping to clarify.

@nicholst and @jbpoline - any basis that you use is often to get at the hemodynamic response. my understanding is that the basis set is the set of functions that you would convolve with a stimulus train. isn't the use of HRF a little superfluous in this context? 

also spm:FIR and fsl:noHRF both can be used without being an HRFBasis. 

perhaps someone can clarify the definition of nidm:HRFBasis ?

afni-rickr:
Yes, the default basis function in afni_proc.py is GAM, however the BLOCK(DUR) functions are possibly more widely used (DUR is the event duration), at least for events lasting longer than 1 or 2 seconds.  There is a fairly large assortment of basis functions, with the most common ones being BLOCK, GAM, TENT (like FIR) and SPMGx.  Duration modulation is also common, where each event gets its own duration, and the BLOCK basis function is convolved as specified per event.

Is there something to specify the event duration?  That might be an attribute of the basis function.

nicholst:
@afni-rickr - thanks for this.  You remind us that there's a lot of complexity.  

But one question: If the user specifies BLOCK(DUR), then what 'smooths' out the block into something that looks like the anticipated response?  Do you additionally have to specify GAM, or is GAM used by default?

About event duration, we aren't (yet) doing proper provenance, i.e. we don't have enough information to reproduce an analysis.  When we get that far we'll definitely have to represent event duration.

afni-rickr:
@nicholst - BLOCK is the actual shape of the curve (as opposed to GAM), which is then convolved with a boxcar of length DUR.  BLOCK is a slightly wider curve than GAM.  Does this boxcar convolution reflect the difference between the HRF and HRB (function vs. basis)?  If not, how do those differ?

I see that you have an FIR list, which AFNI supports via the TENT basis function set.

Maybe @afni-gangc can comment.

nicholst:
Oh, ok.  Then BLOCK is indeed the "HRF" (as it is the thing that is convolved with the stimulus boxcar & generates the predictor).

We decided to use the term "HRF Basis" even if there was just one such response function.

jbpoline:
@afni-rickr : thanks for the explanation - looks like this fits in the category of HRFbasis fairly well.
@satra : Yes, HRFbasis definition would be : "The set of functions to convolve anticipated neural responses to obtain anticipated hemodynamic responses" - or something of the kind.
I would prefer to keep HRFbasis rather than only "basis" : I think it makes it clearer, but no strong feeling either.

nicholst:
@satra - One reason to be specific with HRFBasis is that at some point we might also describe the DriftBasis (DCT in SPM, polynomials in AFNI).

cmaumet:
712db84 adds new terms for afni:BLOCK, afni:FIR. I think we are good to merge this and we can discuss the terms and definitions in a separate issue?

nicholst:
+1 to merge

On Mon, Mar 2, 2015 at 5:39 PM, Camille Maumet <notifications@github.com>
wrote:

> 712db84
> <https://github.com/incf-nidash/nidm/commit/712db84dcb8e46af635d27f3b60a703348d9c1fe>
> adds new terms for afni:BLOCK, afni:FIR. I think we are good to merge this
> and we can discuss the terms and definitions in a separate issue?
>
> —
> <https://github.com/incf-nidash/nidm/pull/248#issuecomment-76746702>.
>



-- 
Thomas Nichols, PhD

Fx,  +44 24 7652 4532

cmaumet:
Just rebased on master.

cmaumet:
Rebased on master +  33f4cf9 and c5df009 update scripts and examples accordingly.

I think this is good to go. I will open separate issues to discuss the definitions.

Issue #249, opened 2015-02-20
NIDM-experiment spec: UML chart is truncated


cmaumet:
As noted by @nicholst in #245, the UML chart is truncated in the right in the NIDM-experiment specification.

cmaumet:
@nicholsn: It looks like you have been working on an updated version of the nidm-experiment UML chart, would you like to update the spec accordingly?

nicholsn:
Sure, I'll try to get this done before the call on monday.

sent via mobile, please excuse brevity and typos

nicholsn:
[Link to current figure](https://drive.google.com/open?id=0B-BLof5_SOh8NGJWeklpLUJPdE0&authuser=0)

nicholsn:
@cmaumet, do you have any instructions on how to generate the spec from the owl file?

cmaumet:
@nicholsn: you can use:
`python nidm/nidm-experiment/scripts/create_expe_specification.py`

nicholsn:
hmm... im getting the following... thoughts?

```
python nidm/nidm-experiment/scripts/create_expe_specification.py
/Users/nolan/Repos/nidm/scripts/OwlReader.py:90: UserWarning: No PROV type for class: nidm:TaskBasedImagingAcquisition
  warnings.warn('No PROV type for class: '+self.graph.qname(class_name))
```

cmaumet:
Hi @nicholsn.  This is just a warning but the html should have been generated properly. Can you check it?

nicholst:
Hi @cmaumet - Here it's still truncated: http://nidm.nidash.org/specs/nidm-experiment_dev.html maybe the source image hasn't been pushed here http://nidm.nidash.org/specs/img/nidm-experiment-uml.png from github?

cmaumet:
@nicholst: sorry this was not clear... I meant the html should have been re-generated in @nicholsn's branch (not pushed yet?) but to update the website at http://nidm.nidash.org, the usual procedure is still needed (cf. [Generate the NIDM website](https://github.com/incf-nidash/nidm/wiki/Generate-the-NIDM-Website)).

nicholst:
Sorry!  I misread @nicholsn  as @nicholst.  Ignore me!

nicholsn:
this was resolved at: http://nidm.nidash.org/specs/nidm-experiment_dev.html

PR #259, opened 2015-03-02
MRG: Updated NIDM Experiment draft UML

nicholsn:
This PR fixes issue #249 and adds a non-truncated version of the draft UML diagram. 

Note: this figure is by no means final and needs further discussion.

The updated figure was push to: http://nidm.nidash.org/specs/nidm-experiment_dev.html


cmaumet:
@nicholsn: if that's okay with you: I will have a quick look tomorrow to make a comparison with the old chart and might send out a few comments.

nicholsn:
Please do, it needs a lot of work still...

sent via mobile, please excuse brevity and typos

cmaumet:
Hi @nicholsn: sorry it took me longer than I thought to get back to this... 

Here are a few comments:
 - Does the "Principal Investigator" role need to be an entity? Or could it be just a term? (i.e. what attributes could we envision for this role?)
 - Maybe "Participant Group" could be at the Modeling level rather than the Session level?
 - Could we keep "Presentation Software" as an agent rather than having a "Software" agent with a role of "Presentation". My point is that I think that being a "Presentation Software" is an intrinsic characteristic: I don't think you could use a "Presentation Software" (like EPrime) for other than Presentation... but maybe people more familiar with those software could comment?

A few typos (from the original chart...):
 - Arrows are reversed for: 
    - "AnatomicalScan wasGeneratedBy AnatomyImagingAcquisition", 
    - "FunctionalSeries wasGeneratedBy TaskBasedImagingAcquisition", 
    - "TaskEvents wasGeneratedBy TaskBasedImagingAcquisition", 
    - "SessionCollection wasGenerateBy SessionAcquisition".

I have noticed that you also changed some naming from the original chart. I have just sent you a PR to make sure that we keep the owl file in sync.

Thanks for this new version of the UML chart! I am happy to merge once those small updates are taken into account.

nicholsn:
thanks for the feedback!

>
>    - Does the "Principal Investigator" role need to be an entity? Or
>    could it be just a term? (i.e. what attributes could we envision for this
>    role?)
>
> ​The principle investigator should actually be a subclass of
`prov:Role`... Attributes may be first name, last name, department, phone
number, orcid id, etc. ​


>
>    - Maybe "Participant Group" could be at the Modeling level rather than
>    the Session level?
>
> ​What if we are sharing a raw anatomical and resting-state dataset​

​with no model level. How would be differentiate groups?​

>
>    - Could we keep "Presentation Software" as an agent rather than having
>    a "Software" agent with a role of "Presentation". My point is that I think
>    that being a "Presentation Software" is an intrinsic characteristic: I
>    don't think you could use a "Presentation Software" (like EPrime) for other
>    than Presentation... but maybe people more familiar with those software
>    could comment?
>
> ​That makes sense, sure.​


>
>
> A few typos (from the original chart...):
>
>    - Arrows are reversed for:
>       - "AnatomicalScan wasGeneratedBy AnatomyImagingAcquisition",
>       - "FunctionalSeries wasGeneratedBy TaskBasedImagingAcquisition",
>       - "TaskEvents wasGeneratedBy TaskBasedImagingAcquisition",
>       - "SessionCollection wasGenerateBy SessionAcquisition".
>
> I have noticed that you also changed some naming from the original chart.
> I have just sent you a PR to make sure that we keep the owl file in sync.
>
​Merged =)​


> Thanks for this new version of the UML chart! I am happy to merge once
> those small updates are taken into account.
>
​Great!​

cmaumet:
> ​The principle investigator should actually be a subclass of `prov:Role`... Attributes may be first name, last name, department, phone number, orcid id, etc. ​

Wouldn't these be attributes of the Person rather than the "Principal Investigator" role?

> ​What if we are sharing a raw anatomical and resting-state dataset​ ​with no model level. How would be differentiate groups?​

Then maybe the "Participant Group" could be at the "Investigation" level?  What I mean is just that the Group a subject is part of is not really an information at the series level (the same subject is always part of the same group (whatever the serie) in a given study). But the distinction between levels is just here to guide the reading of the graph so I am fine as is if you think it is good.

dbkeator:
> - Maybe "Participant Group" could be at the Modeling level rather than the Session level?

Regarding "Participant Group" I vaguely remember talking about the concept of "group" on our last call.  The benefit of having the group defined at the session level was that we could have different group assignments for different sessions.  This is a valid use-case when running a drug study where a subject may be in the placebo group for one session and a drug group for another.  Typically I think subjects get grouped into a placebo or drug group at the time they are enrolled in the investigation (i.e. experiment) but I suppose that's not always the case.  I agree with @nicholsn that having it at the model level would require a model for all sessions which may not be the case for simply sharing structural data.

>​The principle investigator should actually be a subclass of prov:Role... Attributes may be first name, >last  name, department, phone number, orcid id, etc.

I agree with @cmaumet that name, department, etc. are probably attributes of the prov:Person and not the role. I suppose there could be attributes of the prov:Role for PIs but I'm having a hard time coming up with many.  One thing might be the years the PI is PI of the project?  Say for example you have multiple PIs for an investigation where one moves or leaves the site and another PI takes over.  There might be a need for the valid dates a person is serving in the role of PI?  I suppose another example is if you have multiple co-PIs where each is responsible for a different aspect of the investigation.  You'd then want to document which part of the project the particular PI is responsible for (e.g. PI 1 = MRI, PI 2 = EEG, etc...).
 ​

nicholsn:
>
> Typically I think subjects get grouped into a placebo or drug group at the
> time they are enrolled in the investigation (i.e. experiment) but I suppose
> that's not always the case.
>
​The use-case I reference is the NCANDA project, where all enrolled kids
were drug/alcohol naive and​

​the expectation is that some will self select into another group​.

I am thinking about this from a data management perspective, where during
an ongoing study I need to query for information before there are any
models. I'm open to other approaches that meet this requirement, and I
think these annotations can take place at more than one level - modeling
included.

> I agree with @nicholsn <https://github.com/nicholsn> that having it at
> the model level would require a model for all sessions which may not be the
> case for simply sharing structural data.
>
​I don't see why the model level couldn't have a subject group annotation.
Good chance group models will be shared without the session-level info​


>  ​The principle investigator should actually be a subclass of
> prov:Role... Attributes may be first name, >last name, department, phone
> number, orcid id, etc.
>
> I agree with @cmaumet <https://github.com/cmaumet> that name, department,
> etc. are probably attributes of the prov:Person and not the role.
>
​me too after rethinking this - i don't think roles need any further
attributes​ at the instance level

> I suppose there could be attributes of the prov:Role for PIs but I'm
> having a hard time coming up with many. One thing might be the years the PI
> is PI of the project? Say for example you have multiple PIs for an
> investigation where one moves or leaves the site and another PI takes over.
> There might be a need for the valid dates a person is serving in the role
> of PI? I suppose another example is if you have multiple co-PIs where each
> is responsible for a different aspect of the investigation. You'd then want
> to document which part of the project the particular PI is responsible for
> (e.g. PI 1 = MRI, PI 2 = EEG, etc...).
>
​true, but we could postpone modeling this until a use-case presents
itself.​

cmaumet:
> The benefit of having the group defined at the session level was that we could have different group assignments for different sessions. This is a valid use-case when running a drug study where a subject may be in the placebo group for one session and a drug group for another.

> ​The use-case I reference is the NCANDA project, where all enrolled kids were drug/alcohol naive and​ ​the expectation is that some will self select into another group​.

Thanks for the use cases! (I am too biased by the control/patient use-case...)
+1 for keeping the Participant Group at the serie level.

> ​I don't see why the model level couldn't have a subject group annotation.
Good chance group models will be shared without the session-level info​

I had not thought that we might want to ship the different levels separately, thanks for the detailed explanation!

So... I am happy with current version of the UML chart once the arrows are reversed and maybe the roles are set as classes rather than entities (but I agree with @dbkeator  that we might want to add attributes in the future so either way is good for me).

nicholsn:
updated and rebased, should be good to merge

cmaumet:
Thanks @nicholsn! In the [updated UML chart](https://raw.githubusercontent.com/nicholsn/nidm/02ed41bec5674ce3df2871aa4c79274a79ed60fb/doc/content/specs/img/nidm-experiment-uml.png), it looks like the arrows are still reversed and `Presentation` still appears as a role (rather than a `PresentationSoftware` class). Did you forget to upload the latest version? Or am I looking at the wrong file...? Thank you!

nicholsn:
You are looking at the right file - I just didn't update these. Which
arrows are reversed now - I only changed the hadRole arrows. I'll update
the `PresentationSoftware` to a class.

cmaumet:
Thanks @nicholsn. I was looking at the reversed arrows I initially mentioned in [this comment](https://github.com/incf-nidash/nidm/pull/259#issuecomment-77401052).

nicholsn:
thanks! arrows fixed, rebased, and specs updated.

cmaumet:
Great! Seems good to me, let's merge!

Issue #254, opened 2015-02-27
Conversion from provn to turtle


cmaumet:
There are currently a number of pull requests (#248, #241, #252, #253) that do pass the tests when I start them locally from my machine but fail on Travis. 

The test suite fails with a timeout when accessing the ProvStore:

```
.INFO:TestProvStoreLinks:Test: TestProvStoreLinks
INFO:TestProvStoreLinks:    Prov store URL: https://provenance.ecs.soton.ac.uk/store/documents/31582.ttl
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

We are currently linking to the ProvStore in order to convert from ttl to provn (when using Travis. Locally this conversion is done with the java provToolbox). 

I can see two possible solutions to avoid querying the provStore to do the conversion:
1. Start using the python `prov` toolbox to convert from ttl to provn. @satra: do you think we could start beta testing the export on the python `prov` toolbox or shall we wait a bit more?
2. Use the java provToolbox. This would mean that we need to be able to install it as a dependency when starting the tests on Travis CI. I am not sure how to do this given that the rest of the project is in python. Would someone have an example of Travis CI configuration adding a dependency in another programming language?

Thanks in advance for any feedback/tip/comment.

cmaumet:
I have just renamed this issue in "Conversion from provn to turtle". While there is no longer a timeout issue with the ProvStore (cf. #260), we still need to discuss how to handle the provn to turtle conversion in python.

satra:
perhaps i can spend some time on this at the ohbm hackathon. i'll also see if i can get a summer urop to work on this.

cmaumet:
@satra: that would be great! I think @jbpoline also had in mind that others in the NIDM group could contribute (this was in the agenda of the Washington NIDASH TF meeting if I remember correctly).

This is probably something we could discuss on one of the calls.

cmaumet:
For info, the python NIDM-Results exporter now uses rdflib to export to turtle (cf. https://github.com/cmaumet/nidmresults/commit/d3ac863c568ac63cc78a430fa42f65474babb0fd). 

This avoids a dependency on the java provToolbox until the python prov toolbox can export to turtle.

satra:
+1

PR #253, opened 2015-02-26
Specifications: display label rather than URI in the text

cmaumet:
In this pull request, the script to generate the specifications is updated to:
- display the human-readable label stored in `rdfs:label` rather than the `uri` of a term when a label is available.

e.g. for "Contrast Estimation" ([before](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-contrastestimation), [after](http://htmlpreview.github.io/?https://raw.githubusercontent.com/incf-nidash/nidm/ec38fbe8d66aa533bc8262d26c9bbb3f0c9a24fa/doc/content/specs/nidm-results_dev.html#dfn-nidm-contrast-estimation))

PR #252, opened 2015-02-26
Re-use STATO estimation methods

cmaumet:
Following ISA-tools/stato#32, "Ordinary Least Squares estimation", "Weighted Least Squares estimation", "Generalised Least Squares estimation", "Feasible Generalized Least Squares estimation" and "Iteratively Reweighted Least Squares estimation" are now defined in STATO.

This pull request implements the re-use of these terms as a value for `nidm:withEstimationMethod` object property used in the `nidm:ModelParametersEstimation` activity ([current implementation](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-modelparametersestimation), [proposed update preview](http://htmlpreview.github.io/?https://raw.githubusercontent.com/incf-nidash/nidm/8c2daba767b1cc6707677e80b4e4b1e6db5b46e8/doc/content/specs/nidm-results_dev.html#dfn-nidm-modelparametersestimation)).

PR #241, opened 2015-01-21
Automatic testing of specification consistency with owl files

cmaumet:
This pull request implements:
 1. A new test to check that the specifications `nidm-results_dev.html` and `nidm-experiment_dev.html` are in sync with the corresponding owl files.
 2. A couple of small fixes to the NIDM-experiment and NIDM-Results specs, mainly:
  - Un-grey all links at http://nidm.nidash.org/specs/
  - Add the possibility to have several version of the NIDM-Experiment spec (with links);
  - Add links to the development version for both specs.

It would be great if we could merge this before the NIDM-experiment meeting next week. Thanks!

PR #260, opened 2015-03-03
Retry on timeout when calling the ProvStore

cmaumet:
In this pull request:
- we catch timeouts (>5s) on `urlopen` and `read` calls to the ProvStore (and Prov validator)
- retry the call on timeout (up to 15 times).

This fixes issue #254, though we will probably want to avoid calling the prov store in the future (and rather use the prov python toolbox). So I think #254 can remain open.


cmaumet:
I will now merge this pull request as we need it for the tests to pass in #248, #241, #252, #253.

PR #71, opened 2015-03-04
Definition of Cluster

cmaumet:
Cluster definition

Following conversations with @cmaumet and @gllmflndn: If or when we create a cluster entity (instead of cluster size entity), here is a finely crafted definition.

> Cluster
> 
> A collection of voxels that are contiguous.  For example, among all voxels with statistic values above some threshold, a set of voxels that all share at least one face is a cluster.  There are different contiguity criterion; for example in 3D, 6-connectivity specifies that 2 voxels are contiguous if they share a face; 18-connectivity specifies that 2 voxels are contiguous if they share a face or an edge; 26-connectivity specifies that 2 voxels are contiguous if they share a face, an edge or a corner.


cmaumet:
Following our agreement on those definitions, I have implemented our discussion to:
 - Add the definitions of `nidm:Cluster` and `nidm:SignificantCluster` in `nidm-results.owl`
 - Update NIDM-Results examples to replace `nidm:Cluster` by `nidm:SignificantCluster`

If there are no further comments, I think this is good to merge.

nicholst:
+1

cmaumet:
(rebased). Let's merge.

gllmflndn:
I would also suggest to remove "significant" from the entity label and given name - a given cluster may or may not be significant.
http://nidm.nidash.org/specs/nidm-results_110.html#dfn-nidm_0000070
I'd vote for just "cluster" (or "suprathresholdcluster" if something more specific is really needed). Note that the other terms use "cluster" only happily: clusterSizeInVoxels, clusterLabelId, ClusterDefinitionCriteria, ClusterLabelsMap, ...

nicholst:
+1 for @gllmflndn @jbpoline comments on 'significant'.  I think we just want 'cluster' which has been of course defined by some combination of cluster-forming threshold and (possibly) a cluster extent threshold.

satra:
@jbpoline, @gllmflndn - how would this be different from `nidm:cluster`? see prior discussion on this thread.

nicholst:
As detailed in the thread, you wanted `nidm:Cluster` to refer to the generic entity of "clustering" (in it's many many many applications). 

I think @jbpoline and @gllmflndn's point is that you can create (supreathreshold you-know-what-I-mean-in-the-context-of-fmri-spm-fsl-fs) clusters in various arbitrary ways that aren't 'significant'.

I think I could summarise the above conversation by saying we all couldn't get on board for the general definition of 'cluster' and hence we avoided it by creating `nidm:SignificantCluster`.  But we haven't resolved the issue.

I'll agree that we're *not* talking about clusters in the general sense used in biostatistics/bioinformatiocs, and that 'SuprathresholdCluster' is a sharper definition of this.  But there if you try to account for everything under the sun I don't see how we'll ever get anything done.  _In the NIDM namespace_, I don't see any trouble with nidm:cluster or nidm:SignificantCluster, as they will be carefully defined.  What I don't know is how distressed I should be that those terms don't have a good link to BioInfo:Cluster... my sense is that it doesn't create many headaches.

satra:
@nicholst - all the examples i gave were from imaging, not clustering in general (i just don't want us to use fmri stats for our definition of `nidm:cluster`). and even in imaging many aspects of general clusters come into the picture. for example, what if i do kmeans on any imaging data? it's neither suprathreshold nor significant, but it is defined by some criterion.

tiborauer:
How about that?
A collection of voxels that are groupped together based on a certain delineation criterion. For example...

nicholst:
OK... well, I wasn't aware of kmeans output from any existing neuroimaging software, but point taken.  

So then let's agree on a term and consistently use it.  "SuprathresholdCluster" ? And then "SignificantSuprathresholdClusters" ?

khelm:
I'm a bit wary of adding a lot of terms that modify a basic term as terms themselves.  What about clusters that are formed when something is *below* a threshold?  What about a cluster that is defined when the members are equal to a certain value?  It would seem that the parsimonious way to go would be to keep "cluster" generic and add attributes.

nicholst:
I think the problem is that we're trying to accommodate all possible meanings of 'cluster', knowing it is widely used in so many contexts, when we only have a good sense of a few specific use cases ('clusters in a statistic image, formed as a set of voxels above a threshold and satisfying a connectivity criterion' in particular, 'kmeans' in some other setting).  

Hence, if we're going to do this properly we need a *really* *good* definition of "cluster" and then define attributes/special cases of that term.  (A good general definition isn't in STATO, btw, where they only have the notion of 'cluster design', yet another sense!!)

Save having a working group that spends a week to develop a thorough definition of "cluster", I would propose that we _accept_ that neuroimaging's concept of "cluster" is somewhat idiosyncratic relative to other definitions of cluster (which usually involve a complex data analysis procedure like k-means or Ward/hierarchical clustering, not to mention the biostatistical/design concept of a 'cluster' in design and data analysis--e.g. a cluster sampling design randomly samples a whole classroom of students in a school, instead of randomly sampling students independently from the entire student body), and simply define clear and unambiguous terms that serve our needs.  

To that end, that is basically what we've done, but with the slight problem of "significant".  

Returning to 'significant', can I simply propose that we return to @gllmflndn's proposal to drop "significant", and let other attributes of the model (i.e. pvalues) establish the significance or non-significance of the cluster.

satra:
@nicholst - i'm very much -1 on this. let's put this on the discussion for monday. i will try to come up with a more convincing argument as to why it's a bad idea to use such a generic term in the nidm namespace to mean something specific to a sub-domain of brain imaging.

nicholst:
Ok, let's discuss, but to be clear I'm *not* insisting on "cluster" as a
NIDM term, but rather that we only aspire to define the specific terms we
need without getting bogged down in ontological discussions.

On Friday, October 30, 2015, Satrajit Ghosh <notifications@github.com>
wrote:

> @nicholst <https://github.com/nicholst> - i'm very much -1 on this. let's
> put this on the discussion for monday. i will try to come up with a more
> convincing argument as to why it's a bad idea to use such a generic term in
> the nidm namespace to mean something specific to a sub-domain of brain
> imaging.
>
> —
> <https://github.com/incf-nidash/nidm/pull/71#issuecomment-152656596>.
>


-- 
Thomas Nichols, PhD

Fx,  +44 24 7652 4532

satra:
> I'm *not* insisting on "cluster" as a NIDM term, but rather that we only aspire to define the specific terms we need without getting bogged down in ontological discussions.

agreed - it's just that dropping the word `significant` makes the term be called `nidm_cluster` (and even though we are using IDs, this is a case where i do want to be careful with the label).

tiborauer:
One of the advantage of my proposal, that it allows to define a criterion attribute, so we can specify any (below or above threshold, significant size or mass after delineation, or any other common characteristic).

khelm:
I'm -1 on @nicholst 's proposal as well, given that, as @gllmflndn points out that we use"cluster"" in the names of other terms.  Because of that I think we need a generic version of cluster otherwise we'll end up with a lot of specific-case terms involving "cluster" because our base definition is not general enough.  You need generic terms to construct a minimal-branch tree - otherwise you end up with a lot more terms on the same level.

nicholst:
... or we simple acknowledge that we've used ambiguous term, and throughout the term database we change all occurrences of "cluster" to "suprathreshold cluster".

For reference, here are all the occurrences of "cluster" in the term database... they all look to be consistently referring to this one particular notion of 'suprathreshold cluster' neuroimagers usually speak of, and thus we could safely do the search-and-replace.

> has Cluster Labels Map ;
> A property that associates a clusters label map ...
> Property that associates a ConnectivityCriterion with a ClusterDefinitionCriteria. ;
> cluster Label Id ;
> Integer value used in the cluster label map ... of the cluster within ...
> cluster Size In Vertices ;
> Number of vertices that make up the cluster. ;
> cluster Size In Voxels ;
> Number of voxels that make up the cluster. ;
> max Number Of Peaks Per Cluster ;
> Maximum number of peaks to be reported after inference within a cluster. ;
> number Of Significant Clusters ;
> Number of significant clusters found ...
> expected Number Of Clusters ;
> Expected number of clusters in an excursion ...
> expected Number Of Vertices Per Cluster ;
> Expected number of vertices in a cluster under the null hypothesis. ;
> expected Number Of Voxels Per Cluster ;
> Expected number of voxels in a cluster under the null hypothesis. ;
> cluster Size In Resels ;
> Size of cluster measured in resels. ;
> smallest Significant Cluster Size In Vertices FDR 05 ;
> Smallest cluster size in vertices that is significant ...
> smallest Significant Cluster Size In Vertices FWE 05 ;
> Smallest cluster size in vertices that is significant ...
> smallest Significant Cluster Size In Voxels FDR 05 ;
> Smallest cluster size in voxels that is significant ...
> smallest Significant Cluster Size In Voxels FWE 05 ;
> Smallest cluster size in voxels that is significant ...
> Cluster ;
> Cluster Definition Criteria ;
> Set of criterion specified a priori to define the clusters reported after inference (e.g. pixel or voxel connectivity criterion). ;
> Cluster Labels Map ;
> A map ... denotes cluster membership.... Each cluster is denoted ... cluster have...
> ... a lower bound on cluster-sizes ... minimum cluster size in voxels ;
> ... the significant cluster ...
> ... cluster, minimum distance between peaks). ;
> Significant Cluster ;
> A cluster of map elements
>  ... cluster connectivity ...
> Cluster Center Of Gravity ;
> A centre of gravity of the cluster ... the cluster element ... cluster element intensities. (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/Cluster) ;

See also the use of Cluster in the filenames in nidm-results/terms/examples/*Cluster*txt

khelm:
Yes, but the reason for this is that what's been modeled so far is fMRI analysis.  I think the point of contention here is coming down to either viewing terms through only an fMRI lens or something broader - neuroimaging - in which all of the different types of clusters that have been discussed here can arise.  I don't think there needs to be a week of debate on the definition, but I think we first need to decide if we are fMRI-only focused. I would vote for including terms that are applicable to other types.

nicholst:
If we can nail down the appropriate definitions and relationships swiftly I'm all for it. But beyond the current use-case (results reporting for fMRI) is a big world of multifaceted uses that I worry that we couldn't do justice in a finite amount of time.  I don't want to burn any bridges, but we need to admit we can't build all the bridges in the world at once.

jbpoline:
It looks like we are close to a potentially useful definition. For reference, the obi one is 

Term IRI: http://purl.obolibrary.org/obo/OBI_0000251
definition: A data set which is a subset of data that are a similar to each other in some way.

and could be reused for the general meaning ?

+1 for the supra_threshold_cluster

nicholst:
+1 for JB's suggestion of parent cluster term:
http://purl.obolibrary.org/obo/OBI_0000251

BTW, I stumbled across a review paper (Bowman 2014) that nicely summaries
various flavors of clustering (all, except, "suprathreshold clustering" as
it happens :) which might be interesting for reference...

> Cluster analysis can also be used to partition the brain into networks of
voxels or regions that exhibit similar temporal or task-related dynamics.
The statistical literature provides well- established techniques for
performing cluster analysis. Several of these data-driven methods have been
successfully applied in brain imaging, including the following: K-means
approaches (Balslev et al. 2002; Goutte et al. 1999, 2001), fuzzy
clustering methods (Baumgartner et al. 2000; Fadili et al. 2000, 2001;
Somorjai & Jarmasz 2003), hierarchical clustering methods (Bowman & Patel
2004, Bowman et al. 2004, Cordes et al. 2002, Goutte et al. 2001, Stanberry
et al. 2003), a hybrid hierarchical K-means approach (Filzmoser et al.
1999), and dynamical cluster analysis (Baune et al. 1999).Hierarchical
clustering generally begins by treating each brain region (or node) as a
single cluster, calculating the functional distances between all pairs of
brain regions (e.g., fij = 1−ρij , where ρij is the partial correlation
between regions i and j), and iteratively joining the most similar regions
or clusters (recalculating the functional distances at each step). The
process ceases when only a single cluster remains. Clustering procedures
then examine each stage of the resulting tree to determine the optimal
number of clusters, often on the basis of criteria assessing both within-
and between-cluster variability.

Bowman, F. D. (2014). Brain Imaging Analysis. Annual Review of Statistics
and Its Application, 1(1), 61–85.
doi:10.1146/annurev-statistics-022513-115611



On Mon, Nov 2, 2015 at 12:18 AM, Jean-Baptiste Poline <
notifications@github.com> wrote:

> It looks like we are close to a potentially useful definition. For
> reference, the obi one is
>
> Term IRI: http://purl.obolibrary.org/obo/OBI_0000251
> definition: A data set which is a subset of data that are a similar to
> each other in some way.
>
> and could be reused for the general meaning ?
>
> +1 for the supra_threshold_cluster
>
> —
> <https://github.com/incf-nidash/nidm/pull/71#issuecomment-152880086>.
>



-- 
Thomas Nichols, PhD

Fx,  +44 24 7652 4532

jbpoline:
Thanks for the reference- I had not seen it - Dubois could also have cited
a few papers by Guillaume, Ferath and co in the early 2000 :) I think
Guillaume was the first to do a combined functional + spatial clustering,
but I might be wrong

supra_threshold_clustering may be too simple a method to be in this review
:)

cheers
JB

PR #256, opened 2015-03-06
Definition of "ExcursionSet" -> "ExcursionSetMap"

cmaumet:
This is a proposal to update the naming and definition of "ExcursionSet" as follows:

 1. Add a "Map" suffix to be consistent with all other terms representing a Map in nidm (e.g. `ContrastMap`, `ParameterEstimateMap`...).
 2. Update the definition to: "a map in which the set of elements not surviving the inference is set to zero or NaN."

(The current definition of "ExcursionSet" is "Set of map elements surviving a thresholding procedure". )


cmaumet:
I think that we are good here and will now merge this set of updates. 

If you feel this should be further discussed, please feel free to re-open a separate issue.

PR #141, opened 2015-03-06
Definition of "ModelParametersEstimation"

cmaumet:
**Term**: `ModelParametersEstimation`
**Definition**: "The process of estimating the parameters of a general linear model from the available data".
**Original comment**: [link](https://docs.google.com/spreadsheets/d/16pC2cDsdxlzv2CzlNMtStqt5-xHwDEsU6MjZVxWhrE4/edit?usp=sharing)
**Current link**:  [nidm-results.owl/ModelParametersEstimation](https://github.com/incf-nidash/nidm/blob/master/nidm/nidm-results/nidm-results.owl#L2162-L2177)

---

@cmaumet `11:46 25 Apr`
How about re-using the STATO definition: "model parameter estimation is a data transformation part of a regression analysis which applies specific mathematical transformation in order produced estimates on the parameter specified by the model based on the data used as input."
	
@cmaumet `12:11 8 May`
Awaiting STATO reply at ISA-tools/stato#18

### DIFF
--- a/doc/content/specs/nidm-results_dev.html
+++ b/doc/content/specs/nidm-results_dev.html
@@ -562,7 +562,7 @@
             <section id="section-nidm:ModelParametersEstimation"> 
                 <h1 label="nidm:ModelParametersEstimation">nidm:ModelParametersEstimation</h1>
                 <div class="glossary-ref">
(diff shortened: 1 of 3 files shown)


cmaumet:
I think that we are good here and will now merge this set of updates.

If you feel this should be further discussed, please feel free to open a separate issue.

PR #264, opened 2015-03-10
Term README: consider term as uncurated if curation status is missing

cmaumet:
In this pull request, we:
 - include a small update to consider a term as uncurated is its curation status is missing. 
 - update the README file describing the terms accordingly ([current README](https://github.com/incf-nidash/nidm/blob/master/nidm/nidm-results/terms/README.md), [updated README](https://github.com/incf-nidash/nidm/blob/9169227acee06fc36995e53d342689708d72ad96/nidm/nidm-results/terms/README.md)).

### DIFF
--- a/nidm/nidm-results/terms/README.md
+++ b/nidm/nidm-results/terms/README.md
@@ -40,6 +40,26 @@
     <td><b>nidm:SPM: </b>Statistical Parametric Mapping software package for the analysis of neuroimaging data from the FIL Methods Group</td>
 </tr>
 <tr>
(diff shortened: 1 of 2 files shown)


jbpoline:
I was trying to go through the Owlreader.py code and was wondering: would it be possible to have a small owl file in the test directory to make it easier to see what the reader should be returning ?

cmaumet:
(rebased). @jbpoline: I am going to merge this PR in order to have an update list of terms that need curation. But that's a good point: more testing would be needed for the different scripts and classes. And also some refactoring... I am currently focusing on curating the definitions and finishing a first version of the data model but happy to discuss this further in a separate issue if you would like to.

PR #266, opened 2015-03-18
Fix for testing of NIDM-Results SPM exporter

cmaumet:
This pull request includes a couple of fixes to the testing procedure of [NIDM-Results SPM exporter](https://github.com/incf-nidash/nidm-results_spm):
 - take into account owl imports (for STATO terms),
 - better reconciliation across graphs.


cmaumet:
I will now merge this update (only impacting testing and needed at incf-nidash/nidm-results_spm#9).

PR #238, opened 2015-03-18
Use of rdfs:isDefinedBy or iao:"definition" for definitions

cmaumet:
In `nidm-results.owl`, we use [rdfs:isDefinedBy](http://www.w3.org/TR/rdf-schema/#ch_isdefinedby) attribute to specify the definition while in `nidm-experiment.owl`, we use [IAO:"definition" (obo:IAO_0000115)](http://www.ontobee.org/browser/rdf.php?o=IAO&iri=http://purl.obolibrary.org/obo/IAO_0000115). The definition of the latter seems more appropriate but is directly linked with OBO while the former might be more generic. 

Do you know if there is a good practice?


cmaumet:
The use of `obo:definition` (rather than `rdfs:isDefinedBy`) is now implemented. I think this is good to merge.

nicholsn:
sounds good

PR #267, opened 2015-03-19
SPM ex001: DisplayMask removed, dependence model is Spatially Global...

cmaumet:
Following testing at incf-nidash/nidm-results_spm#9, this pull request includes a few fixes in example 001 suggested by @gllmflndn:
 - `DisplayMaskMap` is removed (in the absence of contrast masking, this entity should not be present).
 - Dependence model is set to `nidm:SpatiallyGlobalModel ` instead of `nidm:SpatiallyLocalModel `
 - SPM revision number is set to `12.0`

Additionally:
 - A new script (`recompute_all_ex.py`) is created to re-generate all the turtle examples with a single call.
 - A separate example (005) is created for SPM to describe the minimal set of entities and relations (including `DisplayMaskMap`) expected when contrast masking is performed.


cmaumet:
I am going to merge those changes that only impact testing (and are needed at incf-nidash/nidm-results_spm#9).

Issue #155, opened 2014-08-24
Definition of "MaskMap"


cmaumet:
**Term**: `MaskMap `
**Current definition**: "A map whose value at each location denotes whether the element is of interest for further analysis."
**Original comment**: [link](https://docs.google.com/a/mrn.org/spreadsheets/d/16pC2cDsdxlzv2CzlNMtStqt5-xHwDEsU6MjZVxWhrE4/edit?disco=AAAAAGsYjiI)

---

@khelm `20:26 9 Apr`
This seems overly restrictive. How about: "A matrix whose function is to select a subset of elements in an image or parameter map."? Here it would just be "parameter map".
	
@nicholsn `21:16 18 Apr`
+1
	
@JessicaTurner `20:49 22 Apr`
Are we defining a mask by what it does or how it's built? I think of it as being 1s and 0s as well as being applied to actually mask voxels in or out. But if that's too restrictive...
	
@cmaumet `16:56 23 Apr`
Following Karl and relying on the definition of 'Map' we could have:
"A binary map whose function is to select a subset of elements."

cmaumet:
@JessicaTurner, @khelm, @nicholst: how about:

**MaskMap**: "A binary map in which selected elements (e.g., pixels, voxels, vertices, and faces) have a value of 1 and excluded elements have a value of 0 or NaN."

I think that removing "function" is a bit clearer as a `MaskMap` can be input or output of an Activity.  Adding examples of "elements" might help understanding what we mean (i.e. voxels in a 3D volumetric map).

khelm:
I agree about "function" being problematic if it's an output.  So +1 for the @cmaumet definition.

cmaumet:
I will now close this issue as the definition of **nidm:MaskMap** was discussed and agreed upon in #258.

PR #258, opened 2015-03-19
Definition of "CustomMaskMap", "Mask", "InferenceMask", "DisplayMask" and "SearchSpaceMap"

cmaumet:
This is a proposal to update the definition of "nidm:CustomMaskMap" to:
"A `nidm:MaskMap` defined by the user to restrain the space in which an activity is performed".

current definition: "mask defined by the user to restrain the space in which model fitting is performed."

Not referring explicitly to "model fitting" would allow us to re-use this term at different level of the analysis pipeline.


cmaumet:
This is now implemented as follows (changes from the previous version are in bold):

- *nidm:MaskMap* "A binary map in which selected elements (e.g., pixels, voxels, vertices, and faces) have a value of 1 and excluded elements have a value of 0 or NaN." (re-using definition proposed in #155). -- This term is parent of all other masks --.

 - *nidm:DisplayMaskMap*: "A mask defined by the user as an input of inference that changes the voxels displayed, but does not alter the search space in which the inference is performed **e.g. it does not alter voxel-wise corrected p-values** (**this mask is called** “contrast masking” in both FSL & SPM).

 - <i><b>nidm:UserDefinedMaskMap</i></b>: “A mask defined by the user to restrain the space in which an activity is performed “ -- replace nidm:CustomMask and nidm:InferenceMask.

 - *nidm:AnalysisMaskMap*: "A mask generated by a model parameter estimation activity that restricts the space in which contrast estimation and inference are performed." -- replace current nidm:MaskMap

 - <i>nidm:SearchSpace<b>Mask</b>Map</i>: "A mask in which the inference was performed." -- replace nidm:SearchSpaceMap. 

Please let me know if that's what you were thinking.

khelm:
+1  for the term names
and
I would alter: nidm:DisplayMaskMap: "A mask defined by the user as an input of inference that ..."
to
nidm:DisplayMaskMap: "A mask defined by the user as an input of an inference activity that..."
I'm not sure I get the e.g. that you added.  The definition talks about not altering the search space and the e.g. talks about not altering p-values.  There seems to be an intermediate step in between the two and it might help to be a bit more explicit.  That will make the definition longer however.  Can we live without the e.g.?

nicholst:
I have only one beef with these definitions.  

`nidm:AnalysisMaskMap` has the wiggly definition "mask ... that restricts the space".  Is anyone else bothered by this?  I mean, in *all* these definitions, the one that most people understand as "the mask" by default, is the "analysis mask", which a user (not a party to this discussion) would define as "The exact set of voxels in which the mass univariate model is estimated".  This intuitive description precisely describes the output of the model estimation activity.  However, as I understand it, we don't like such as sharp definition, as it is inconsistent with the definition with the input to inference (where this "analysis mask" must be subsequently intersected with the `nidm:UserDefinedMaskMap`, aka SVC map).

If this doesn't bother anyone, then it's fine, but if anyone else shares my reservations, let's sharpen the *one* definition of *mask* that a gentle user might intuit.

tiborauer:
I also vote for @nicholst definition. I think, "generated by" and "defined by" are already specified in the attributes.

cmaumet:
@nicholst: I understand your concern. How about combining both definitions to get:
 - *nidm:AnalysisMaskMap*: "A mask representing the exact set of voxels in which the mass univariate model is estimated. This mask is further used to restrict the space in which contrast estimation and inference are performed."

?

@khelm: what happened is that I tried to re-use some of the definition we initially agreed on in #157... How about (re-using a larger part of the initial definition proposed by @nicholst):
 - *nidm:DisplayMaskMap*: "A mask defined by the user as an input of **an** inference **activity** that changes the voxels displayed, but does not alter the search space **used for multiple-testing correction** e.g. it does not alter voxel-wise corrected p-values (this mask is called “contrast masking” in both FSL & SPM).

?

nicholst:
@cmaumet - OK, I like the first sentence.  But I the spirit of "less is more" do we really even need the 2nd sentence? Won't that be clear in the inference activity?

tiborauer:
nidm:AnalysisMaskMap: Is also think it is obvious. Especially, because  you can estimate contrast and inference only where the model is estimated.

nidm:DisplayMaskMap: Again, I am not sure whether "defined by" should be included. 

I was thinking: Is not it possible to simply use a general definition: "A MaskMap, which represent the exact set of voxels in which the corresponding activity is performed."
In this case, AnalysisMaskMap, SearchSpaceMaskMap, DisplayMaskMap would be simply instances (not subclasses) of MaskMap. The provenance (defined/generated, used) would be obvious from the provanence model itself. 
Another option/idea I was thinkg about is to have "dynamic" definition/description (if possible), which would use the provenance model itself to contsruct self. If it is not implemented yet, would not it be a good idea?

cmaumet:
@nicholst: +1 for a simplified definition of *nidm:AnalysisMaskMap* (i.e. "A mask representing the exact set of voxels in which the mass univariate model is estimated.") and to give more details when defining *nidm:Inference*.

@tiborauer: I understand your proposal but again I think this is a philosophical question that was addressed when discussing the role of the different NIDM components. Have you looked at my [detailed reply to your initial comment](https://github.com/incf-nidash/nidm/pull/258#issuecomment-78097503)?

cmaumet:
I have just thought that one alternative could be to use an attribute `isUserDefined` to specify whether or not the mask was defined by the user. In that case we could have a limited number of `MaskMap` classes (@tiborauer and @nicholst) as follows:

- **nidm:BinaryMap** "A map in which all values are 0, 1 or NaN. Usually, selected elements (e.g., pixels, voxels, vertices, and faces) have a value of 1 and excluded elements (i.e. the background) have a value of 0 or NaN. -- This term will be parent of all other mask maps defined below --.

  - **nidm:MaskMap** "A binary map representing the exact set of elements (e.g., pixels, voxels, vertices, and faces) in which an activity was performed (e.g. the mask map generated by the model parameter estimation activity represents the exact set of voxels in which the mass univariate model was estimated) and/or restraining the space in which an activity was performed (e.g. the mask map used by inference). -- currently nidm:CustomMask (with `nidm:isUserDefined = true`) and nidm:MaskMap (with `nidm:isUserDefined = true`).  
    - **nidm:SearchSpaceMaskMap**: "A mask map representing the set of elements in which the inference was performed." -- currently nidm:SearchSpaceMask. -- SearchSpaceMask is child of MaskMap --

  - **nidm:DisplayMaskMap**: "A binary map used by an activity that changed the voxels displayed, but did not alter the space in which the activity was performed (e.g. at the level of inference, this mask is called "contrast masking" in both FSL & SPM and does not alter voxel-wise corrected p-values) -- currently nidm:DisplayMask.  
  
 
Using a `userDefined` attribute is similar to what we have done for [nidm:ExtentThreshold](http://nidm.nidash.org/specs/nidm-results_dev.html#section-nidm:ExtentThreshold) and would allow us to separate function (DisplayMask versus Mask) from origin (user-defined or not) as suggested by @khelm.

How do you feel about this proposed update?

tiborauer:
+1
I like any change which uses modelling instead of hard-coding. :)

nicholst:
+++1 for @cmaumet's last proposal.  I especially like BinaryMap, because I sometimes deal with image data that take the form of binary images (lesion analysis in Multiple Sclerosis and stroke), and it would be nice to know that image data of that form is already the model somewhere.

khelm:
+1 from me too.  It seems like a general solution and one that makes it easy to build upon.

nicholst:
Quick question @cmaumet: is **nidm:DisplayMaskMap** a child of **nidm:MaskMap**? Your indenting made it look like it was a child of **nidm:BinaryMap**.

cmaumet:
@nicholst: I was thinking that **nidm:DisplayMaskMap** would be child of **nidm:BinaryMap** (and not of **nidm:MaskMap**) because the definition of **nidm:MaskMap** does not apply for **nidm:DisplayMaskMap**. Do you think this is fine?

tiborauer:
@cmaumet: I have read your detailed response just now. It somewhow eluded my attention. :)

I thought, NIDM-Results is a continuation of NIDM-Workflow, and the model estimation is the border(?). I just take a look and you are righ! There is no SVC activity in NIDM-Results. I took it granted. Is not it missing?
According to what you say, NIDM-Workflow may overlap/include NIDM-Results. Is that the plan? I guess I have to take a look on those slides. Unfortunately, I can do it only from home.

So, actually, it is the NIDM-Result, which is more mechanical (more explicitly hard-coded)!

nicholst:
@cmaumet - So I guess if 'display' isn't an activity, then **nidm:DisplayMaskMap** doesn't belong under **nidm:MaskMap**, but intuitively **nidm:DisplayMaskMap** *feels* like "A binary map representing the exact set of elements *used *for* *visualisation* and/or restraining the space in which *visualisation* occurred." no?

cmaumet:
@nicholst: I completely concur with your comment: if we had a "display" activity then the "DisplayMaskMap" could be modelled as a "MaskMap" used by "display". Though, I am not sure we want to go that road... If we were to create a new "display" (or SVC) activity we would then have to create a first excursion set (without contrast masking) generated by "inference" and used by "display" and then a second one (with contrast masking) generated by display. While this would be more thorough in terms of modelling, I think it would add a level of complexity that is probably not desirable in NIDM-Results... And if we do so, we will probably also want to have a separate activity to create the clusters (and peaks). Please let me know what are your thoughts on this.

nicholst:
OK, @cmaumet, I'm fine with this.  Ontological consistency strikes again!

cmaumet:
Implemented as agreed and ready to go!

PR #269, opened 2015-03-20
Search volume in voxels in nidm namespace

cmaumet:
In this pull request, discussed with @nicholst, we:
 - replace `spm:searchVolumeInVoxels` and `fsl:searchVolumeInVoxels` by a common term: `nidm:searchVolumeInVoxels`.

As the concept of *search volume measured in voxels* is the same across neuroimaging software, it seems that having a single term in the `nidm` namespace would be more appropriate (than having two separate terms for SPM and FSL).


cmaumet:
(Rebased on master)

nicholst:
+1 ... I'd say merge this puppy to support the rapid work needed on terms...

tiborauer:
+1
Simplification is good! :D

cmaumet:
Thank you and let's merge!

PR #296, opened 2015-04-08
Use of alphanumeric identifiers in NIDM-Results

cmaumet:
This pull request introduces the use of alphanumeric identifiers (e.g. `nidm_NIDM_0000231`) in place of semantic identifiers (e.g. `nidm:StatisticMap`) as discussed on a special NIDM call on Dec 18th 2014 (cf. [slides](https://drive.google.com/open?id=0ByYytkxqR-v0MXV6LU4tVGlMMHM&authuser=0)).

An example of how the specification is updated:
 - [nidm:'Contrast Map' (updated version with alphanumeric ids)](http://htmlpreview.github.io/?https://raw.githubusercontent.com/incf-nidash/nidm/f689291ba0087806f30d9202223bb96a2aac270a/doc/content/specs/nidm-results_dev.html#dfn-nidm_0000002)
 - [nidm:ContrastMap (old version with semantic ids)](http://htmlpreview.github.io/?https://raw.githubusercontent.com/incf-nidash/nidm/23f214700c78bfe24dc7532b9798e77bbe9cd1d9/doc/content/specs/nidm-results_dev.html#dfn-nidm-contrastmap)

(Please note, that, for now, this update is only applied to the `nidm:` namespace (as discussions are still ongoing w.r.t defining which terms should appear in other namespace, e.g. `spm:`, `fsl:`, `afni:`).)

It would be great to get some feedback on this. Thank you in advance!


nicholst:
As before, +1 for alpha numeric ids.

I really like the documentation, with the numeric ID appearing when you hover over a term.

cmaumet:
We discussed this on April 13th.

cmaumet:
Following discussions on NIDM call:
 - diamonds have been replaced by direct links for `<dfn>` tags.
 - After having a closer look at the second point. It looks like we would need to modify Respec javascript as currently the "id" is generated from the "title" field (and the "title" field is displayed in the "Definition list"). I have implemented a quick work-around but I am not really satisfied with it as it means re-writing all the links (instead of relying on Respec)... So I have no solution for this right... 

New preview:  [nidm:'Contrast Map' (updated version with alphanumeric ids)](http://htmlpreview.github.io/?https://raw.githubusercontent.com/incf-nidash/nidm/5281299ed246064b827b58eb83025b3793d516dd/doc/content/specs/nidm-results_dev.html#dfn-nidm_0000002)

I think this is good to go (keeping in mind the issue with the "Definition list")!

nicholst:
Looks great; just note that the 'Definition list' includes semantic identifiers.

cmaumet:
Small fix: d2b58cd removes two remaining diamonds (pointed out by @nicholst).

cmaumet:
I am going to merge this as agreed so that we can start building new pull requests based on alphanumeric identifiers.

Issue #35, opened 2014-04-04
What namespace to use for specifications developed by the INCF Neuroimaging Task Force


nicholsn:
This is a continuation from a conversation between @satra, @cmaumet, and @nicholsn at #34.

This issue concerns deciding on a namespace for specifications developed by the INCF Neuroimaging Task Force and the Derived Data Working Group that supports content negotiation (see [COOLURIS](http://www.w3.org/TR/2008/NOTE-cooluris-20081203/#choosing)).

In particular, this is to choose a namespace for NI-DM Object Model specifications and corresponding vocabulary serializations.

Currently there are a few options:
- http://nidm.nidash.org/terms#
- http://www.nidash.org/nidm#
- http://www.incf.org/ns/nidm#
- http://www.incf.org/ns/nidash#
- http://purl.org/nidash#
- http://purl.org/nidash/nidm#
- ...

Notes:
- purl.org is a free redirection site many ontology/vocabulary developers use
- the namespaces with only a "nidash" reference a "meta" page pointing to data model and any other non-data model resources developed by the task force.
- the namespaces with a "nidm" reference data model specific terms (e.g., object models)

PR #237, opened 2015-04-09
purl of incf URI for nidm namespace

cmaumet:
This follows suggestion by @nicholsn in #236, to use a purl URI for the NIDM-Experiment namespace: http://purl.org/nidash/nidm/experiment. Previously (cf. #35) we had agreed on using an INCF URI http://www.incf.org/ns/nidash/nidm. 

This raises two questions:
 1. Do we want to use purl-based (http://purl.org/nidash/xxx) instead of INCF-based (http://www.incf.org/ns/nidash/xxx) namespaces?
 2. Do we want a separate namespace for each NIDM-component (e.g. http://purl.org/nidash/nidm/experiment, http://purl.org/nidash/nidm/results...) or a common NIDM namespace for all components (e.g. http://purl.org/nidash/nidm/)?

Note: In both cases (purl or INCF URI), we will have to make the namespace URI dereferenceable and display the terms and definitions at that location (which is not currently the case for the INCF-based namespace URI).


cmaumet:
As agreed, use of purl namespace (e.g. `nidm:` is http://purl.org/nidash/nidm#) is now implemented for NIDM-Results and NIDM-experiment. Let's merge!

PR #257, opened 2015-04-10
Definition of "center of gravity"

cmaumet:
This is a proposal to expand the definition of `fsl:CenterOfGravity`:

"Centre of gravity of the cluster, equivalent to the concept of Centre Of Gravity for a object with distributed mass, where intensity substitutes for mass in this case. **The coordinate of the centre of gravity is the weighted average of the coordinates by the intensities within the cluster** (definition from http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/Cluster)"

(The expanded part of the definition also comes from http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/Cluster).


cmaumet:
Now implemented and ready to merge.

PR #261, opened 2015-04-10
1st level fMRI: drifts modelling

cmaumet:
[amended to take into account English corrections suggested below by @nicholst].

This is a proposal discussed with @nicholst, to include information about drift models in first-level fMRI design matrices (which is part of the check-list in ["Guidelines for reporting an fMRI study." Poldrack et al. 2008](http://dx.doi.org/10.1016/j.neuroimage.2007.11.048).):
- a new `nidm:hasDriftModel` relation from `nidm:DesignMatrix` to `nidm:DriftModel`; 
- a new `nidm:DriftModel` class with children 
  - `spm:DCTDriftModel`: Discrete Cosine Transform basis added to regression model.  
  - `fsl:GaussianHighPassDriftModel`: Locally weighted running-line smoother with Gaussian weights, applied to and subtracted from the data and each column of the design matrix.  
  -  `afni:LegendrePolynomialDriftModel`: Legendre orthogonal polynomial basis added to regression model. 
- new software-specific attributes to describe the parameters of each approach:
  - `spm:driftCutoffFrequency`: cut off frequency/periodicity (slowest frequency in the basis),
  - `fsl:driftHighPassFilterCutoff`: "longest temporal period allowed"
  - `afni:driftBasisOrder`: Basis order; 1=linear, 2=quadratic, etc.

e.g. with SPM:
pr 261- drift model in design matrix

The precise naming and definition of those new terms would be discussed in a separate issue, once we agree on the model.

How do you feel about this update?


cmaumet:
Now implemented, to be consistent with the rest of `nidm-results.owl` I have kept the `nidm:hasDriftModel` Object Property. I am going to merge and open a separate issue to discuss the definitions. If you think we need more thoughts on how to model drift models please feel free to reopen an issue.

PR #255, opened 2015-03-19
Re-use STATO terms in nidm Map definitions (e.g. ContrastMap)

cmaumet:
To make the best of the terms defined in STATO, how about updating our Map definitions to link to stato:

- **nidm:ContrastMap**: A map whose value at each element is a [`stato:"contrast estimate"`](http://purl.obolibrary.org/obo/STATO_0000384).
- **nidm:ParameterEstimateMap**: A map whose value at each element is a [`stato:"model parameter estimate"`](http://purl.obolibrary.org/obo/STATO_0000144).
- **nidm:StatisticMap**: A map whose value at each element is a [`stato:"statistic"`](http://purl.obolibrary.org/obo/STATO_0000039).

PR #287, opened 2015-04-13
Definition of "fsl:ContrastVarianceMap" (current status: uncurated)

cmaumet:
This issue is open to discuss the definition of **fsl:ContrastVarianceMap**.

###### Proposal
Given that the "contrast variance" is a generic concept, I would suggest to move this term to the `nidm` namespace and, similarly to what we have done for other *Map* terms in #255, how about:
 - **nidm:ContrastVarianceMap**: A map whose value at each element is a contrast variance (i.e. the square of a [`stato:"standard error of a contrast estimate"`](http://purl.obolibrary.org/obo/STATO_0000385)).


cmaumet:
Thanks! Now implemented and ready to go.

PR #300, opened 2015-04-14
Search volume in units (and in vertices) in nidm namespace

cmaumet:
Similarly to what was done in #269 for "searchVolumeInVoxels", could we:
 - replace `spm:searchVolumeInUnits` by a generic term in the `nidm` namespace: `nidm:searchVolumeInUnits`
 - replace `spm:searchVolumeInVertices` by a generic term in the `nidm` namespace: `nidm:searchVolumeInVertices`;

?


cmaumet:
Thank you! Now implemented. Let's wait for the tests to pass and then merge.

PR #293, opened 2015-04-14
Definition of (remaining) object properties: statisticType, withEstimationMethod, hasMIP

cmaumet:
Similarly to what was agreed on for other object properties, I think we could go for:

 - **nidm:statisticType**: Property that associates a  [`stato:"statistic"`](http://purl.obolibrary.org/obo/STATO_0000039) with a StatisticMap or a ContrastWeights entity.
 - **nidm:withEstimationMethod**: Property that associates a  [`stato:"model parameter estimation"`](http://purl.obolibrary.org/obo/STATO_0000119) method with a model parameter estimation. 
 - ~~spm:~~**nidm:hasMaximumIntensityProjection**: Property that associates an image of the maximum intensity projection with an excursion set map.


cmaumet:
Thank you! This is now implemented. Let's wait for the tests to pass and then merge.

PR #286, opened 2015-04-15
Definition of "fsl:ClusterMaximumStatistic" (current status: uncurated)

cmaumet:
This issue is open to discuss and curate the definition of **fsl:ClusterMaximumStatistic**.

###### Proposal
Similarly to what we have done for *nidm:Inference* and *nidm:ContrastWeights* in order to ease querying, I would propose to use an attribute *nidm:isClusterMaximum* instead of having a specific class *fsl:ClusterMaximumStatistic*. This would lead to:
<pre>
niiri:peak_0001_1 a prov:Entity , nidm:Peak ;
    rdfs:label "Peak 0001_1" ;
    <b>nidm:isClusterMaximum "true"^^xsd:boolean ;</b>
    ... other attributes ...
    prov:wasDerivedFrom niiri:significant_cluster_0001 .
</pre>
instead of (current version):
<pre>
niiri:peak_0001_1 a prov:Entity , nidm:Peak, <b>fsl:ClusterMaximumStatistic</b> ;
    rdfs:label "Peak 0001_1" ;
    ... other attributes ...
    prov:wasDerivedFrom niiri:significant_cluster_0001 .
</pre>

The definition could be as follows:
 - **nidm:isClusterMaximum**: Binary flag defining whether the peak corresponds to the absolute maximum of the cluster (true).

Note: this information is currently only stored for FSL (where it is part of the standard result display) but as this is a generic concept, I think we could use the `nidm` namespace as proposed above.


cmaumet:
Thank you! Implemented and ready to go.

PR #283, opened 2015-04-15
Definition of "Peak" (current status: metadata complete)

cmaumet:
This issue is open to curate the definition of **nidm:Peak**:

###### Proposal
Similarly to what was agreed on for [*nidm:SignificantCluster*](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-significantcluster), how about:
 - **nidm:Peak**: map element (e.g., pixels, voxels, vertices, and faces) which is a local maximum in the significant cluster and complies with the peak definition criteria.

?

###### Current definition
 - *nidm:Peak*: Statistic defined at the peak-level in an excursion set. FIXME (now Peak instead of PeakStatistic).

###### Term usage
A `nidm:Peak` was derived from a `nidm:SignificantCluster` which itself was derived from a `nidm:ExcursionSetMap` (see also [example of usage of Peak](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-peak) in the spec).


cmaumet:
Thank you! Now implemented and ready to go.

PR #273, opened 2015-03-24
Definition of "ExtentThreshold" (metadata complete)

cmaumet:
This issue is open to curate the definition of **nidm:ExtentThreshold**.

###### Current definition
- *nidm:ExtentThreshold*: "A numerical value that establishes a bound on a range of cluster-sizes. / [from extentThresh:] Minimum cluster size used when thresholding a statistic image 5voxels"

###### Term usage
A `nidm:ExtentThreshold` is used an an input of `nidm:Inference` (see also [ExtentThreshold in the spec](http://nidm.nidash.org/specs/nidm-results_dev.html#section-nidm:ExtentThreshold)). 

 - E.g. of `nidm:ExtentThreshold` entity for a cluster-wise threshold of p<0.05:
```
niiri:extent_threshold_id a prov:Entity , nidm:ExtentThreshold ;
	rdfs:label "Extent Threshold: p<0.05 (FWE)" ;
	nidm:userSpecifiedThresholdType "p-value FWE"^^xsd:string ;
	nidm:pValueFWER "0.05"^^xsd:float .
```

 - E.g. of `nidm:ExtentThreshold` entity with a cluster-wise threshold of k>=0:
```
niiri:extent_threshold_id a prov:Entity , nidm:ExtentThreshold ;
	rdfs:label "Extent Threshold: k>=0" ;
	nidm:clusterSizeInVoxels "0"^^xsd:int ;
	spm:clusterSizeInResels "0"^^xsd:float ;
	nidm:pValueUncorrected "1"^^xsd:float ;
	nidm:pValueFWER "1"^^xsd:float .
```

-----

###### Proposal

How about adding a small extension to the current definition:
- *nidm:ExtentThreshold*: A numerical value that establishes a bound on a range of cluster-sizes **and can be specified by the user in terms of FWER-corrected p-value, uncorrected p-value or minimum cluster size in voxels**.

? Would you have alternative suggestions or comments? Thank you!

PR #280, opened 2015-04-15
Definition of "HeightThreshold" (current status: pending final vetting)

cmaumet:
This issue is open to curate the definition of **nidm:HeightThreshold**.

###### Current definition
- *nidm:HeightThreshold*: "A numerical value that establishes a bound on a range of voxelwise or vertex-wise defined statistic."

###### Term usage
A `nidm:HeightThreshold` is used an an input of `nidm:Inference` (see also [example of usage of HeightThreshold in the spec](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-heightthreshold)). 

###### Proposal

Similarly to what was recently agreed for *nidm:ExtentThreshold* (cf. #273), how about:
- *nidm:HeightThreshold*: **A numerical value that establishes a bound on a range of statistic values and can be specified by the user in terms of FWER-corrected p-value, uncorrected p-value or minimum statistic value.**

?


cmaumet:
Thank you! Now implemented and ready to go.

PR #271, opened 2015-03-31
Definition of `clusterLabelId` (pending final vetting)

cmaumet:
This issue is open to curate the definition of **nidm:clusterLabelId**.

###### Current definition
- *nidm:clusterLabelId*: Integer associated with a particular cluster as specified in the clusterLabelsMap. 

###### Term usage
`nidm:clusterLabelId` is an attribute in entities of type `nidm:SignificantCluster`, e.g.:
<pre>
niiri:significant_cluster_0002 a prov:Entity , nidm:SignificantCluster ;
	rdfs:label "Significant Cluster: 0002" ;
	<b>nidm:clusterLabelId "2"^^xsd:int ;</b>
	...
	prov:wasDerivedFrom niiri:excursion_set_map_id .

niiri:excursion_set_map_id a prov:Entity , nidm:ExcursionSetMap ;
	rdfs:label "Excursion Set Map" ;
	...
	nidm:hasClusterLabelsMap niiri:cluster_label_map_id .
</pre>

-----

###### Proposal
I wonder if something like:
 - *nidm:clusterLabelId*: An integer value used in the cluster label map to identify the location of the cluster. 

would be more explicit?

Are you happy with the original definition? Would you have alternative suggestions or comments? Thank you!

PR #282, opened 2015-04-15
Definition of "ClusterLabelsMap" (current status: uncurated)

cmaumet:
This issue is open to curate the definition of **nidm:ClusterLabelsMap** and its corresponding object property **nidm:hasClusterLabelsMap**:

###### Proposal
How about:
 - **nidm:ClusterLabelsMap**: Map whose value at each location denotes cluster membership within the excursion set. Each cluster is denoted by a different integer.
 - **nidm:hasClusterLabelsMap**: Property that associates a clusters label map to the excursion set.

?

Note:
 - The proposed definition of *nidm:hasClusterLabelsMap* is similar to definitions used for other object properties (e.g. [nidm:hasConnectivityCriterion](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-hasconnectivitycriterion), [nidm:hasErrorDependence](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-haserrordependence))
 - Adding "within the excursion set" to the definition of *nidm:ClusterLabelsMap* is consistent with what was agreed upon for *nidm:clusterLabelId* in #271).

###### Current definitions
 - *nidm:ClusterLabelsMap*: \<undefined\>
 - *nidm:hasClusterLabelsMap*: A map whose value at each location denotes cluster membership. Each cluster is denoted by a different integer.

###### Term usage
A `nidm:hasClusterLabelsMap` is an attribute of `nidm:ExcursionSetMap`, used to associate a `nidm:ClusterLabelsMap` with the excursion set (see also [example of usage of ClusterLabelsMap](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-clusterlabelsmap) and [ExcursionSetMap](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-excursionsetmap) in the spec).

### DIFF
--- a/doc/content/specs/nidm-results_dev.html
+++ b/doc/content/specs/nidm-results_dev.html
@@ -1504,7 +1504,7 @@ niiri:cluster_definition_criteria_id a prov:Entity , nidm_ClusterDefinitionCrite
             <section id="section-nidm:'Cluster Labels Map'"> 
                 <h1 label="NIDM_0000008">nidm:'Cluster Labels Map'</h1>
                 <div class="glossary-ref">
-                    A <a title="NIDM_0000008"><dfn title="NIDM_0000008">nidm:'Cluster Labels Map'</dfn></a> is . <a title="NIDM_0000008">nidm:'Cluster Labels Map'</a> is a prov:Entity. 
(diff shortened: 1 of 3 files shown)


cmaumet:
Implemented with a minor fix: "at each location" was replaced by "at each element (e.g., pixel, voxel, vertex, or face)" for consistency with other "Map" definitions.

This is also ready to go! Thank you!

PR #146, opened 2015-04-15
Definition of "dimensionsInVoxels"

cmaumet:
**Term**: `dimensionsInVoxels `
**Current definition**: "Dimensions of some N-dimensional data."
**Original comment**: [link](https://docs.google.com/spreadsheets/d/16pC2cDsdxlzv2CzlNMtStqt5-xHwDEsU6MjZVxWhrE4/edit?disco=AAAAAGsYcCg)

---

@khelm `20:08 9 Apr`
How about: Dimensions of a data set matrix.
	
@nicholsn `22:37 18 Apr`
It would be helpful for attribute definitions to give an idea of the expected values. Something like:

"Dataset attribute describing the dimensions of said dataset as an array of integers, where the first integer of the array corresponds to the first dimension, the second integer to the second dimension, and so on."
	
@JessicaTurner `21:22 21 Apr`
I think "matrix" may be too limiting. But on the other hand, isn't "dimensions" a scalar? (e.g. 3, or 2, or 4?)
--oh never mind, I saw the example. Nolan's definition definitely fits the example.
	
@cmaumet `17:34 23 Apr`
+1 for Nolan's definition. 

[more comments on "Map" term not reported here]


cmaumet:
Thank you! Now ready for release.

PR #292, opened 2015-04-15
Definition of "nidm:numberOfClusters" (current status: uncurated)

cmaumet:
###### Proposal
How do you feel about:
 - **nidm:numberOfClusters**: number of significant clusters found by the inference activity. 

?

###### Current usage
*nidm:numberOfClusters* is an attribute of a *nidm:ExcursionSetMap*, cf. [example of usage](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-excursionsetmap) in the spec.


cmaumet:
So, I think we are good with the definition and this is now implemented.

Regarding the naming, I agree with @khelm that "numberOfSignificantClusters" is more consistent with the definition, and as our "clusters" are now called "significant clusters" (since #71) I think it makes sense to add "significant". 

As we are now using alphanumeric identifiers, this means updating the label to "number Of Significant Clusters". I have implemented this as well. 

I think this is good to go unless there are further comments?

nicholst:
If it's in the definition, sure, I agree that `numberOfSignificantClusters` and more precise... +1

cmaumet:
(Rebase on master). Thank you! I think this is good to go.

PR #276, opened 2015-04-15
Definition of "ContrastEstimation"

cmaumet:
This issue is open to curate the definition of **nidm:ContrastEstimation**.

###### Proposal
Following the creation of *stato:'contrast estimation'* (cf. ISA-tools/stato#37), and similarly to what was done for *nidm:'ModelParametersEstimation'* in #141, how about:

 - **nidm:ContrastEstimation**: The process of performing a stato:"Contrast Estimation" at each element of a map.

?

###### Current definition
- *nidm:ContrastEstimation*: The process of estimating a contrast from the estimated parameters of statistical model. 

###### Term usage
`nidm:ContrastEstimation` is a type of Activity, cf. [example in the spec](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-contrast-estimation).


cmaumet:
This is now implemented and good to go?

nicholst:
Looks ready to me.

khelm:
+1

Issue #114, opened 2014-07-08
NIDM: Perspective and scope


satra:
NIDM is intended to capture a dataflow - not just the data, not just the workflow, not just "results"- but how data are generated and transformed. The abstractions made on this dataflow capture specific components that might be relevant to applications at hand (e.g., meta-analysis). But parts of the same dataflow may be relevant to other applications (e.g., predictive analytics). Such abstractions were intended to be built on top of the underlying dataflow.

The reason NIDM is built on PROV-DM is to capture the dataflow - the complete provenance of how things come to be and who and what software are involved. Users and developers may want to select and/or transform components of this dataflow into specific object models for consumption.

There are three areas that require additional consideration.
1. Names: Names, particularly extensions, are relevant in brain imaging because there is no consistent file format. From a content perspective a nii.gz is a transformation of an nii file. Their content hashes are going to be different, but most metadata would be the same (filetype, offsets would be notable exceptions). However, every developer needs to know the filetype (more than name) in order to write to disk. This is not even looking into retrospective data that may be stored in img/hdr pairs (analyze or nifti), brik/head, minc/minc2 and other file formats. _Proposal:_ Add filetype metadata to each entity that is a file. 
2. Disconnects: By creating new data elements that are not explicitly connected to the old entities (for example, using prov:alternateOf, prov:specializationOf) we are, as @tiborauer points out, performing actual transformations instead of describing it. From a provenance perspective any disconnect creates the very problems we were trying to address in the first place. _Proposal:_ Add provenance to new entities being created and keep link to original data.
3. Linked Open Data: In the context of Linked Open Data, information is passed around through metadata and the actual binary blobs may be accessed or downloaded based on this metadata, or could be reframed to be downloadable chunks (e.g., allen brain institute data on hi-res images, current considerations for accessing grayordinates in the HCP data, etc.,.). One goal of this Web connected world is to deliver packets that contain information rather than files in particular. While this is still down the road, it should be kept in mind.

In the context of the NIDM-Results project, there is an effort to describe part of the analysis with a tool, integrate this tool in the software, and create new entities. This is creating some odd scopes. Here are some suggestions.
1. If the tool creates new things, we should be tracking their provenance.
2. We really should not be concerned with the richness of information. Our abstractions/tools built on top of this should simplify the output. No human should really be reading things beyond our examples.
3. NIDM data descriptors, workflows and results are all part of a more global dataflow. We really should not be forgetting that. 
[...]

PR #285, opened 2015-04-15
Definition of "nidm:DataScaling" (replacing "nidm:Data")

cmaumet:
This issue is open to curate **nidm:Data** and its attributes.

###### Proposal
We created the *nidm:Data* class a while ago, as a placeholder for the input data and were planning to provide a more precise description in a future version. Since then we had a lot of discussions on the relationship between NIDM-Results and NIDM-Workflows (e.g. in #114 and [slides](https://docs.google.com/presentation/d/1W4l_xhxE4LpKkDhQouBVab0H-W1E-299wm50ZT-j4B0/edit?usp=sharing)) and I think now would be a good time to rethink about this entity. 

As the *nidm:Data* entity is only storing information about data scaling and not data itself (cf. [example of usage](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-data)), I would like to propose to rename it to *nidm:DataScaling* with the following definition:
 - **nidm:DataScaling**: Scaling applied to the data before parameter estimation, including specification of the target intensity.

In the future, we could connect with the actual Data entities created as part of NIDM-Workflow.

###### Current definition (for reference)
*nidm:Data* is currently defined as "A collection or single item of factual information, derived from measurement or research, from which conclusions may be drawn", borrowing the definition from [ncit:data](http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl#C25474).

### DIFF
--- a/doc/content/specs/nidm-results_dev.html
+++ b/doc/content/specs/nidm-results_dev.html
@@ -486,12 +486,12 @@ niiri:coordinate_space_id_1 a prov:Entity , nidm_CoordinateSpace: ;
                             </tr>
                 
                         <tr>
-                            <td><a title="NIDM_0000018">nidm:'Data'</a>
+                            <td><a title="NIDM_0000018">nidm:'Data Scaling'</a>
                             </td>
                     
                                 <td rowspan="8" style="text-align: center;">NIDM-Results Types<br/>                                 (PROV Entity)</td>
                         
-                                <td><a title="NIDM_0000018">nidm:'Data'</a></td>
+                                <td><a title="NIDM_0000018">nidm:'Data Scaling'</a></td>
                             </tr>
                 
                         <tr>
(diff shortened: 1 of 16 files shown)


cmaumet:
Thank you! The definitions are implemented as follows:
 - **nidm:DataScaling**: Scaling applied to the data before parameter estimation, including specification of the target intensity.
 - **nidm:grandMeanScaling**: Binary flag defining whether the data was scaled (true for scaled). Specifically, "grand mean scaling" refers to multiplication of every voxel in every scan by a common (or session-specific) value.
 - **nidm:targetIntensity**: Value to which the grand mean of the Data was scaled (applies only if grand mean scaling is true).

If there are no further comments, I think this is good to go.

cmaumet:
(Rebased on master). Thank you for your comments! This is ready to go.

Issue #70, opened 2014-06-14
Create a Grand mean map entity


cmaumet:
Discussed with @nicholst, @gllmflndn, (and before with @jbpoline, @chrisfilo and @tiborauer):
- [x] Create a `grand mean map` entity as an output of `Parameter Model estimation`.
- [x] Move `medianIntensity` to be an attribute of `grand mean map`

PR #288, opened 2015-04-15
Definition of "GrandMeanMap"

cmaumet:
This issue is open to curate the definition of **nidm:GrandMeanMap** and its attributes.

###### Proposal
Similarly to what we have done for other *Map* terms in #255, how about:
 - **nidm:GrandMeanMap**: A map whose value at each element is the mean of the means (i.e. the grand mean) of the input data to the activity that generated this map.
 
###### Current usage
Cf. [example of usage](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-grandmeanmap) in the specs.

@jbpoline, @nicholst: would you have suggestions to improve the definition of "grand mean"? Thank you!

### DIFF
--- a/doc/content/specs/nidm-results_dev.html
+++ b/doc/content/specs/nidm-results_dev.html
@@ -940,7 +940,7 @@ niiri:error_model_id a prov:Entity , nidm_ErrorModel: ;
             <section id="section-nidm:'Grand Mean Map'"> 
                 <h1 label="NIDM_0000033">nidm:'Grand Mean Map'</h1>
                 <div class="glossary-ref">
-                    A <a title="NIDM_0000033"><dfn title="NIDM_0000033">nidm:'Grand Mean Map'</dfn></a> is . <a title="NIDM_0000033">nidm:'Grand Mean Map'</a> is a prov:Entity. 
+                    A <a title="NIDM_0000033"><dfn title="NIDM_0000033">nidm:'Grand Mean Map'</dfn></a> is a map whose value at each element (e.g., pixel, voxel, vertex, or face) is the mean over all observations of that element in the input maps (after any scaling of those input maps). <a title="NIDM_0000033">nidm:'Grand Mean Map'</a> is a prov:Entity. 
                 </div>
                 <p></p>
                 <div class="attributes" id="attributes-nidm:'Grand Mean Map'"> A <a title="NIDM_0000033">nidm:'Grand Mean Map'</a> has attributes:
--- a/nidm/nidm-results/terms/README.md
+++ b/nidm/nidm-results/terms/README.md
@@ -132,11 +132,6 @@ Discussed with @afni-rickr in <a href="https://github.com/incf-nidash/nidm/pull/
 </tr>
 <tr>
     <td><img src="../../../doc/content/specs/img/red.png?raw=true"/>  </td>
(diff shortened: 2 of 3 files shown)


cmaumet:
I have just implemented @nicholst's latest proposal and think it's ready to go. 

@khelm, @tiborauer: are you happy with this final version?

tiborauer:
+1

khelm:
I hate to say this at this late date, but I'm not sure I understand the process that this definition is supposed to represent.  I get that Grand Mean scaling is scaling each element's value by the mean over all brain voxels in all volumes in the session.  Is that what we're talking about here? It doesn't seem to be...  (Sorry!)

nicholst:
Hi @khelm:  This is something that isn't presently implemented in SPM, but is in FSL (mean_func).  For fMRI, it is simply the average, over time, at each voxel.  In a 2nd level analysis, it is the mean over subjects, at each voxel.

Would the definition be clearer with this addition?

- `nidm:GrandMeanMap`: A map whose value at each element is the mean over all observations **at that element** in the input maps (after any scaling of those input maps).

The "after any scaling" is a subtle but important point:  There is always some scaling done to the raw image data; either a single multiplicative factor just to change the mean/median to some standard value (e.g. 10,000 or 100), or, as in VBM or PET sometimes, a scan-specific multiplier is used.  The parenthetical phrase, thus, is there to establish that it is not the mean of the raw input data, but rather the input data after any selected scaling operations.

Does this help?

cmaumet:
@khelm: Just to add on @nicholst's answer, we introduced the *nidm:GrandMeanMap* following discussions at HBM 2014 (#70) to serve as a diagnostic for data scaling. 

Looking at the values in the *nidm:GrandMeanMap* (and in particular the value of its *nidm:maskedMedian* attribute) is useful as a diagnostic to check that the grand mean data scaling was done as expected (i.e. values in the *nidm:GrandMeanMap* should be close to the specified value of *nidm:targetIntensity*).

khelm:
Could we change "at that element" to "of that element"? (Clearly we've all been spending a lot of time on this if it's coming down to this!)  Otherwise +1 for the new definition from @nicholst

jbpoline:
I like 

- nidm:GrandMeanMap: A map whose value at each element is the mean over all observations at that element in the input maps (after any scaling of those input maps).

but I am not sure about the scaling : it may mean that each observation has been scaled across element (eg voxels of one subject / time point) or it may mean that there is an overall scaling across elements and observations.  I suppose this does not matter in this definition - and would vet it as it is.

khelm:
@jbpoline yes, I agree about the possibility of different scalings, but I think that's covered by saying "any"?

jbpoline:
@khelm : I agree - it is covered by the any, which makes it very broad.  

would this cover the case where there is no scaling ? : 

- nidm:GrandMeanMap: A map whose value at each element is the mean over all observations at that element in the input maps (*possibly* after any scaling of those input maps).

cmaumet:
@jbpoline: I wonder if adding "possibly" might be misleading in the sense that, if there was any scaling, then the GrandMeanMap *must* be computed on scaled data... so I have not implemented your last comment. If you think this should be discussed further, please feel free to open a new issue and we can add this in our next release.

cmaumet:
(Rebase on master). Thank you for the discussions! I think this is ready to go.

PR #281, opened 2015-04-15
Definitions of HRF terms (current status: uncurated)

cmaumet:
Following #248, this issue is open to discuss the definitions of the terms used to model the hemodynamic response function (HRF):

First, let's define the parent class and its associated object property:
 - **nidm:HemodynamicResponseFunctionBasis**: A basis function, part of a set of functions that is convolved with the anticipated neural responses to obtain the anticipated hemodynamic responses. (adapted from the definition suggested by @jbpoline in #248).
 - **nidm:hasHRFBasis**: Hemodynamic response function basis used to model the anticipated hemodynamic response.

Definitions for children of *nidm:HemodynamicResponseFunctionBasis*:
- **nidm:HemodynamicResponseFunction**: Hemodynamic response function basis that can on its own be used to represent the idealised hemodynamic response function.
 - **nidm:GammaDifferenceHRF**: Hemodynamic response function which is a linear combination of two Gamma functions.
   - **fsl:GammaDifferenceHRF**:  Hemodynamic response function which is a mixture of two Gamma functions - a standard positive function at normal lag, and a small, delayed, inverted Gamma, which attempts to model the late undershoot (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html). This is the default in FSL.
    - **spm:GammaDifferenceHRF**: Hemodynamic response function which is a linear combination two Gamma functions and is denoted by "canonical HRF" in SPM. This is the default in SPM.
 - **nidm:GammaHRF**: Hemodynamic response function which is a Gamma function.
   - **afni:GammaHRF**: Hemodynamic response function which is a pre-defined Gamma function. This is the default in AFNI (-GAM option to 3ddeconvolve, or implied via afni_proc.py).
    - ~~**fsl:GammaHRF**: Hemodynamic response function basis that is a Gamma variate (in fact a normalisation of the probability density function of the Gamma function); width and lag can be altered (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html).~~
 - **nidm:GaussianHRF** ~~**fsl:GaussianHRF**~~: Hemodynamic response function which is a gaussian kernel.
 - ~~fsl:NoHRF~~: 

- **nidm:HemodynamicResponseFunctionDerivative**: Hemodynamic response function basis which is the derivative of an hemodynamic response function.
 - **fsl:TemporalDerivative**: Hemodynamic response function basis that is the derivative with respect to time of the FSL's Gamma Difference heamodynamic response function.
 - **spm:DispersionDerivative**: Hemodynamic response function basis that is the derivative with respect to spatial dispersion of the SPM's Gamma Difference heamodynamic response function.
 - **spm:TemporalDerivative**: Hemodynamic response function basis that is the derivative with respect to time of the SPM's Gamma Difference heamodynamic response function.

- **nidm:FiniteImpulseResponseHRB** Hemodynamic response function basis which is a set of Finite impulse response (FIR) filters, with FIR the convolution kernel is represented as a set of discrete fixed-width "impulses" (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html), denoted by TENT in AFNI.
 - ~~afni:FiniteImpulseResponseHRB~~
 - ~~fsl:FiniteImpulseResponseHRB~~
 - ~~spm:FiniteImpulseResponseHRB~~
- **nidm:GammaHRB**: Hemodynamic response function basis which is a set of Gamma functions, also denoted by "FLOBS" in FSL.
 - ~~**fsl:GammaHRB**: Hemodynamic response function basis which is a set of Gamma variates of different widths and lags, also denoted by "FLOBS" in FSL (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html).~~
 - ~~**spm:GammaHRB**: Hemodynamic response function basis which is a set of Gamma functions; window length and order can be altered.~~

- **nidm:FourierHRB**: Hemodynamic response function basis which is a set of Fourier basis.
  - ~~**spm:FourierHRB**: Hemodynamic response function basis which is a set of Fourier basis; window length and order can be altered.~~

- **nidm:CustomHRB**: Hemodynamic response function basis which is a customised set of basis functions.
  - ~~**fsl:CustomHRB**: Hemodynamic response function basis which is a customised set of basis functions, setup in a plain text file with one column for each basis function, sampled at the temporal resolution of 0.05s  (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html).~~
- **nidm:SineHRB** ~~**fsl:SineHRB**~~: Hemodynamic response function basis which is a set of Sine waves of differing frequencies (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html).

(Not sure about **afni:BLOCK**)

As presented above, I am proposing to reserve spm- fsl- and afni-specific terms for pre-defined functions (e.g. SPM canonical HRF, FSL double gamma, afni GAM), these will be defined as individuals.

How do you feel about those definitions?

### DIFF
--- a/doc/content/specs/nidm-results_dev.html
+++ b/doc/content/specs/nidm-results_dev.html
@@ -611,7 +611,7 @@ niiri:data_id a prov:Entity , nidm_DataScaling: , prov:Collection ;
                     <li><span class="attribute" id="nidm:'Design Matrix'.label">rdfs:label</span>: an                     <em class="rfc2119" title="OPTIONAL">OPTIONAL</em> human readable description                     of the nidm:'Design Matrix'.</li>
                      
                         <li><a title="NIDM_0000088"><dfn title="NIDM_0000088">nidm:'has Drift Model'</dfn></a></span>: an <em class="rfc2119" title="OPTIONAL">OPTIONAL</em> a property that associates a drift model to a design matrix (only used for first-level fMRI experiments).(range <a title="NIDM_0000087">nidm:'Drift Model'</a> such as <a title="LegendrePolynomialDriftModel">afni:'AFNI's Legendre Polinomial Drift Model'</a>, <a title="GaussianRunningLineDriftModel">fsl:'FSL's Gaussian Running Line Drift Model'</a>, <a title="DCTDriftModel">spm:'SPM's DCT Drift Model'</a>)</li> 
-                        <li><a title="NIDM_0000102"><dfn title="NIDM_0000102">nidm:'has HRF Basis'</dfn></a></span>: an <em class="rfc2119" title="OPTIONAL">OPTIONAL</em> .(range <a title="NIDM_0000036">nidm:'Hemodynamic Response Function Basis'</a> such as <a title="BLOCK">afni:BLOCK</a>, <a title="CustomHRB">fsl:CustomHRB</a>, <a title="SineHRB">fsl:SineHRB</a>, <a title="NIDM_0000028">nidm:'Finite Impulse Response HRB'</a>, <a title="NIDM_0000030">nidm:'Gamma HRB'</a>, <a title="NIDM_0000037">nidm:'Hemodynamic Response Function Derivative'</a>, <a title="NIDM_0000035">nidm:'Hemodynamic Response Function'</a>, <a title="FourierHRB">spm:FourierHRB</a>)</li> 
+                        <li><a title="NIDM_0000102"><dfn title="NIDM_0000102">nidm:'has HRF Basis'</dfn></a></span>: an <em class="rfc2119" title="OPTIONAL">OPTIONAL</em> a property that associates a set of functions that, when convolved with the anticipated neural responses, yield a set of regressors to model the anticipated hemodynamic responses in a design matrix.(range <a title="NIDM_0000036">nidm:'Convolution Basis Set'</a> such as <a title="BLOCK">afni:BLOCK</a>, <a title="CustomHRB">nidm:'Custom Basis Set'</a>, <a title="NIDM_0000028">nidm:'Finite Impulse Response Basis Set'</a>, <a title="FourierHRB">nidm:'Fourier Basis Set'</a>, <a title="NIDM_0000030">nidm:'Gamma Basis Set'</a>, <a title="NIDM_0000037">nidm:'Hemodynamic Response Function Derivative'</a>, <a title="NIDM_0000035">nidm:'Hemodynamic Response Function'</a>, <a title="LinearSplineHRB">nidm:'Linear Spline Basis Set'</a>, <a title="SineHRB">nidm:'Sine Basis Set'</a>)</li> 
                         <li><a title="hasfMRIDesign"><dfn title="hasfMRIDesign">nidm:'has fMRI Design'</dfn></a></span>: an <em class="rfc2119" title="OPTIONAL">OPTIONAL</em> a property that associates an fMRI design to a design matrix in first-level fMRI analyses.(range <a title="fMRIDesign">nidm:'fMRI Design Type'</a>)</li> 
                         <li><a title="regressorNames"><dfn title="regressorNames">nidm:'regressor Names'</dfn></a></span>: an <em class="rfc2119" title="OPTIONAL">OPTIONAL</em> a list of abstract names associated with each column of the design matrix (e.g. ["motor_left", "motor_right"]).(range <a title="string" href ="http://www.w3.org/2001/XMLSchema#string">xsd:string</a>)</li>        
                 </ul>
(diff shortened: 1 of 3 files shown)


cmaumet:
Current state of our discussions is now implemented. (Note for later: before merging check that all terms are assigned an alphanumeric id).

gllmflndn:
Random thoughts:
* Hemodynamic[**Impulse**ResponseFunction](http://en.wikipedia.org/wiki/Impulse_response)
* [**Partial**Derivative](http://en.wikipedia.org/wiki/Partial_derivative)
* spm:GammaDifferenceHRF: in the literature, you find more references to *Double Gamma HRF* than *Gamma Difference HRF*. I think I actually prefer spm:CanonicalHRF, and leave the actual construction as a linear combination of two gamma probability density functions in the definition.

nicholst:
@cmaumet : I wanted to stress that the parameters of the Gamma are *not* estimated, but fixed. So, yes, I'm fine with replacing "pre-defined" with "fixed".

@gllmflndn 
 - Fine with HemodynamicImpulseResponseFunction if we implement it consistently.
 - Partial derivative is indeed more correct, but do we ever have a naked "Derivative" term?  Is "TemporalDerivative" or "DispersionDerivative" ever ambiguous?
 - I invented the "Gamma Difference" term, as the Double Gamma is completely ambiguous... two gammas added, multiplied, convolved???  We did toss "CanonicalHRF" back and forth, but it's only "Canonical" for SPM (FSL's "Canonical" HRF, if there is one, is the "Single Gamma").  Hence we decided to drop it.

The good thing is that, with alphanumeric ID's, all of these labels and definitions can be easily changed in the future!

satra:
@nicholst - started sweating right here!

> The good thing is that, with alphanumeric ID's, all of these labels and definitions can be easily changed in the future!

only if the concept doesn't change!

nicholst:
Stop sweating @satra! :) I was specifically thinking of just this type of case where there is no confusion about the concept – – a "double gamma" HRF "canoncal"ised in SPM @ similarly implemented in FSL & AFNI
– – but we struggle with a good label for it.

jbpoline:
+1 for the proposals. One tiny thing if we want to be consistant : 
obo:IAO_0000115 "Hemodynamic response function which is a fixed difference of two gamma functions and is denoted by \"canonical HRF\" in SPM. This is the default in SPM" 

should be:

obo:IAO_0000115 "Hemodynamic response function which is a fixed difference of two gamma *distribution* functions and is denoted by \"canonical HRF\" in SPM. This is the default in SPM" 

As I am trying to review things a quick question: for suggested changes that should not be controversial such as the one above, do you want a PR or comments here ?

jbpoline:
I think I would like to change:

nidm:HemodynamicResponseFunctionBasis: A basis function, part of a set of functions that is convolved with the anticipated neural responses to obtain the anticipated hemodynamic responses. 

To:
nidm:HemodynamicResponseFunctionBasis: A set of functions that convolved with the anticipated neural responses yield a set of regressors to model the anticipated hemodynamic responses.

Is the term regressor ok here ?

cmaumet:
+1 for @gllmflndn's proposal to rename **spm:GammaDifferenceHRF** in **spm:CanonicalHRF**, I think it makes it more explicit (and we still have the generic type **nidm:GammaDifferenceHRF**).

+1 for @jbpoline's updates

@gllmflndn: I did not get what is your proposal regarding "HemodynamicImpulseResponseFunction", are you suggesting to replace "FiniteImpulseResponseHRB" by this term? I liked to have a common suffix (i.e. "HRB") for terms representing a set of basis (as opposed to a single basis). What do you think?

I am happy with "partial derivative" or "derivative", what do other think?

@jbpoline: both comments or PR (on top of the original one, e.g. here to cmaumet/nidm/def_hrf) are good.

nicholst:
+1 for having both **spm:CanonicalHRF** & **nidm:GammaDifferenceHRF**

@jbpoline: Thanks for catching the inconsistency with "gamma function"/"gamma distribution function", but for consistency (and correctness w.r.t. statistical terminology) let's make this "gamma probability distribution function".  ( @cmaumet: Please search for all instances of "gamma", and where referring to the 'gamma function', make sure it reads "gamma probability distribution function"; I thought I had caught all instances, but obviously missed some.)

+1 for @jbpoline's update, with following fix
 - `nidm:HemodynamicResponseFunctionBasis`: A set of functions that**, when** convolved with the anticipated neural responses**,** yield a set of regressors to model the anticipated hemodynamic responses.

gllmflndn:
@cmaumet: I think you can forget my comments about *impulse* and *partial*: maybe they can only appear in the definition text.

What about "gamma probability **density** function"?

More generally, I see it was already discussed in #248 but I'm still unsure whether "hemodynamic" should appear that often in terms and definitions. Why not nidm:TemporalBasisSet or nidm:ConvolutionBasisSet or nidm:ConvolutionKernel, and have HemodynamicResponseFunction only in relation to the Canonical/Double Gamma/Difference Gamma options? There's nothing hemodynamic per se to a Fourier, Gamma, FIR basis set. See for example:
  http://dx.doi.org/10.1016/j.jneumeth.2009.08.005
  http://dx.doi.org/10.1016/j.neuroimage.2012.09.014

nicholst:
A [Google Fight](http://www.googlefight.com/index.php?word1=probability-distribution-function&word2=probability-density-function) says "probability distribution function" is about twice as common, but I'm not really fussed... could have either distribution or density.

I like the generality that @gllmflndn suggests with the kernel/convolution terms.  Do we need to make it clear we're talking about time, or just a single dimension?  Obviously convolution is used in 2- and 3-D, but in different contexts.

cmaumet:
@gllmflndn: I see your point and I agree having more generic types would lead to a more robust representation. As we would then get out of the area of neuroimaging, this also means that we would have to look into other ontologies to check which terms we can re-use (as we probably don't want to define such a general ontology by ourselves...). Unfortunately, I will not have enough time before the end of this week to look into this. So, unless someone would like to take the lead on this, we will have to wait for the next release (1.1?). If you are fine with this suggestion, I suggest we open a separate issue to keep a track of this discussion.

gllmflndn:
[Wikipedia] (http://en.wikipedia.org/wiki/Gamma_distribution#Probability_density_function), [MathWorks] (http://www.mathworks.com/help/stats/gampdf.html) and [Wolfram](http://reference.wolfram.com/language/ref/GammaDistribution.html) seem to use *density*.

Is it also too late to rename some of the terms into nidm:FiniteImpulseResponseBasisSet, nidm:GammaBasisSet, nidm:FourierBasisSet, nidm:CustomBasisSet or should I stop worrying thanks to the alphanumeric IDs?

nicholst:
... and my CRC book of Statistics Tables & Formulae also uses _density_.

cmaumet:
So I would say +1 for density (not implemented yet).

@gllmflndn: the issue is that if we remove "HRF"/"HRB" in all those terms we are doing more than changing a label (we are talking about a different concept). If we remove "HRF" and "HRB" in the term names, then it would not make much sense to keep them defined as children of *nidm:HemodynamicResponseFunctionBasis*. This parent class is needed as a range for *nidm:hasHRFBasis*. 

Was your suggestion to rename *nidm:HemodynamicResponseFunctionBasis* in *nidm:ConvolutionBasisSet*? If so, I think that could work (though we would need to check for existing terms later on). If this is what you had in mind, would you mind drafting a definition? I will try to implement the update as I understood it.

cmaumet:
@gllmflndn, your update is now implemented as follows. Could you please let me know if this is what you had in mind?

 - **nidm:ConvolutionBasisSet**: \<undefined\>
 - **nidm:hasHRFBasis**: A property that associates a set of functions that, when convolved with the anticipated neural responses, yield a set of regressors to model the anticipated hemodynamic responses with a design matrix.

Definitions for children of *nidm:HemodynamicResponseFunctionBasis*:
- **nidm:HemodynamicResponseFunction**: *-- same as before --*
  - *-- same as before --*

- **nidm:HemodynamicResponseFunctionDerivative**: *-- same as before --*
  - *-- same as before --*

- ~~nidm:FiniteImpulseResponseHRB~~ **nidm:FiniteImpulseResponseBasisSet** ~~Hemodynamic response function basis which is a s~~ Set of Finite impulse response (FIR) filters, with FIR the convolution kernel is represented as a set of discrete fixed-width "impulses" (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html), denoted by TENT in AFNI.

- ~~nidm:GammaHRB~~ **nidm:GammaBasisSet**: ~~Hemodynamic response function basis which is a s~~ Set of gamma probability distribution functions, also denoted by "FLOBS" in FSL.

- ~~nidm:FourierHRB~~ **nidm:FourierBasisSet**: ~~Hemodynamic response function basis which is a s~~ Set of Fourier basis.

- ~~nidm:CustomHRB~~ **nidm:CustomBasisSet**: ~~Hemodynamic response function basis which is a c~~ Customised set of basis functions.

- ~~nidm:SineHRB~~ **nidm:SineBasisSet**: ~~Hemodynamic response function basis which is a s~~ Set of Sine waves of differing frequencies (definition adapted from http://fsl.fmrib.ox.ac.uk/fsl/fsl4.0/feat5/detail.html).

- ~~nidm:LinearSplineHRB~~ **nidm:LinearSplineBasisSet**: ~~Hemodynamic response function basis which is a~~ A Linear (order 1) spline, providing an estimate that is continuous over time (in contrast to a FIR basis, which is discontinuous between each time bin). This is called TENT in AFNI's 3dDeconvolve program.

Are others also happy with this update?

jbpoline:
+1 for density

so 
- nidm:GammaBasisSet: Hemodynamic response function basis which is a s Set of gamma probability ~~distribution~~ density functions, also denoted by "FLOBS" in FSL.

jbpoline:
and +1 for @gllmflndn 's suggestions

nicholst:
Important fix for `nidm:GammaBasisSet`:  This isn't FLOBS!  FSL's FLOBS
uses a Bayesian prior to generate lots of feasible HRFs and then does a PCA
on them (the result comes out looking just like SPM's informed basis set,
as it happens!)

So, just delete this: ~~Set of gamma probability distribution functions,
also denoted by "FLOBS" in FSL.~~

On Fri, Apr 17, 2015 at 6:07 AM, Jean-Baptiste Poline <
notifications@github.com> wrote:

> and +1 for @gllmflndn <https://github.com/gllmflndn> 's suggestions
>
> —
> <https://github.com/incf-nidash/nidm/pull/281#issuecomment-93893891>.
>



-- 
Thomas Nichols, PhD

Fx,  +44 24 7652 4532

jbpoline:
Just a thought : I wonder if we shouldn't leave out reference to specific software implementation (eg TENT here) to make it more universal. I'm not sure it adds much and if we want to do a mapping this should probably not be part of the definition.

nicholst:
The reason to include the software specific information is just one of accessibility, so that when a software user or developer sees these definitions they know what is what. So I am for retaining these software references until we have some richer form of documentation to link our terms to the software. 

Thomas Nichols, PhD
Principal Research Fellow, Head of Neuroimaging Statistics

Web: http://go.warwick.ac.uk/tenichols
Phone, Stats: +44 24761 51086, WMG: +44 24761 50752
Fax:  +44 24 7652 4532

tiborauer:
One small grammatical change:
* **nidm:hasHRFBasis**: A property that associates a set of functions that, when convolved with the anticipated neural responses, yield a set of regressors to model the anticipated hemodynamic responses **in** a design matrix.

tiborauer:
How about modelling the implementation instead of "hard-coding" it in the definition?

nicholst:
@tiborauer, indeed, as I understand it, the proper provence approach is to model the gory detail of each implementation.  The high-wire act we're doing with NIDM Results, however, is trying to define concepts that are equivalent-as-possible across software, in support of the meta-analysis use case.  

Even if something only lives in one software, we want to define and name it in an objective and more canonical way than in the software itself, and hence "nidm:LinearSplineBasisSet" which is pretty clearly understood outside of AFNI-land, as opposed to "afni:TENTBasisSet".

nicholst:
+1 for the grammar fix.

tiborauer:
@nicholst, how about having both? Using the generally known name and deifnition with an attribute "hasImplementation". E.g.: 
 * nidm:LinearSplineBasisSet 
  * nidm:hasImplementation afni:TENTBasisSet
(hasImplementation may have already been define in another namespace...)

nicholst:
This is certainly what NIDM Workflow is supposed to encode, but I don't
know enough about the workings of PROV to know if this makes sense...
Let's see what others say.

satra:
+1 for @gllmflndn suggestions

@nicholst - i think it's good to have conceptual terms. when we will get to encoding actual algorithms is way beyond the immediate horizon, but perhaps will come about as some link with nidm:Workflow, some github repo, etc.,. but before we get links to such implementation details, SPM, FSL, AFNI, and others will need to make the source code available at a URL accessible repo such as GitHub.

I think for the timebeing this terminology hierarchy/graph or "ontology" is our replacement for such provenance. i think we can simply use subclasses as we have done in other areas to denote specialization of concept w.r.t. a particular implementation.

jbpoline:
@satra : so what is your recommendation for the definition ? leave it conceptual and have an afni namespace for "TENT"  that will "subclass" the general definition? or leave it as it is?

khelm:
Just wondering about @cmaumet 's comment above about removing all references to HRF from the definitions of terms that are children of nidm:HemodynamicResponseFunctionBasis...

+1 for the hasImplementation: idea from @tiborauer if that's workable

satra:
@jbpoline - yes, my idea was to have the afni TENT definition subclass the nidm one.

jbpoline:
+1

cmaumet:
Thank you for these discussions! 

The implementation is now rebased on master and include the last minor fixes suggested by @tiborauer and @nicholst.

Regarding the modelling of software implementation, we went back and forth on this point... We started with @satra's latest suggestion to have sub-classes to represent the software-specific implementations (e.g. in my first post in this issue we had `fsl:FiniteImpulseResponseHRB`, `spm:FiniteImpulseResponseHRB` as children of `nidm:FiniteImpulseResponseHRB`). To be consistent with what we have done with other parts of the model (e.g. the [Error Model](http://htmlpreview.github.io/?https://raw.githubusercontent.com/cmaumet/nidm/fe746fea3727f3433dd5357bfbb0be296a678737/doc/content/specs/nidm-results_dev.html#section-nidm:'Error Model')), my initial suggestion was to delete the subclasses and only use a generic nidm class instead. This is where we ended up adding "this is xxx in SPM, yyy in FSL, zzz in AFNI" in the definitions to make them more accessible. 

I agree that it would be even better to have a machine-readable attribute to encode which implementation was used as suggested by @tiborauer. But I think this need to be further discussed in order to be implemented consistently across NIDM-Results (and not just for the HRF). So for now, I have kept the "this is xxx in SPM/FSL/AFNI" in the definitions. Issue #313 is open to follow up on the implementation discussion.

cmaumet:
@nicholst: I have removed the mention to FLOBS in `nidm:GammaHRB`. Before merging this, I think we need to create a term to cover the FLOBS case in FSL. How about:

- **nidm:BayesianPriorHRFwithPCA**: Hemodynamic response function basis which uses a Bayesian prior to generate a set of possible hemodynamic response functions and then does a principal component analysis on them. This is denoted by "FLOBS" in FSL.

?

This term will not be ready for release (as it would need further review) but we could have it with a "pending final vetting" status if you agree on this.

nicholst:
@cmaumet  - regarding FLOBS, I think there are various details that define FLOBS, and BayesianPriorHRFwithPCA captures some of them, but given the limited use of FLOBS (and how there aren't analgous things in other software packages) whether it merits a parent 'nidm:' term?  

I just have a few moments now, but later I can help sharpen a general description of FLOBS, but it's such a definate (and non-obvious) thing, it feels like it should just be something in the fsl: namespace.  Is it a problem to only define something in the fsl: name space and then just link it to 'HRFs' (or whatever the name is) in nidm.

cmaumet:
Thank you @nicholst. It's probably better to keep FLOBS out of the release for now (i.e. link to the parent class when FLOBS is called as you suggested). I have open issue #315 to follow up on how to model FLOBS so that we can include it in our next release.

cmaumet:
Thank you for the discussions! I am going to merge this update. 

Please note that `afni:BLOCK` and `nidm:ConvolutionBasisSet` are not yet ready for release and will need further discussions (cf. [term README](https://github.com/cmaumet/nidm/blob/adaf9bd03ebf2f04c322c6cd9fd08ceb463afaa7/nidm/nidm-results/terms/README.md)).

PR #214, opened 2015-01-21
Refining the definition of FWHM, and defining FSL's equivalent measure

cmaumet:
Following discussions with @cmaumet, motivated by development of the FSL results model, we realised that our term definitions for noise spatial smoothness (aka FWHM) were a bit ropey.  This issue proposes some revisions to the SPM smoothness terms *and* proposes some new terms for the FSL parameters.

The present terms `spm:noiseFWHMInUnits`, `spm:noiseFWHMInVertices` and `spm:noiseFWHMInVoxels` are ambiguous because they make no mention of *space*, and as such one could take FHWM to be a description of the univariate noise variance!  

Current descriptions, as pulled from [terms README](https://github.com/incf-nidash/nidm/blob/master/nidm/nidm-results/terms/README.md): 

* **spm:noiseFWHMInUnits**: Estimated Full Width at Half Maximum of the noise distribution in world units.
* **spm:noiseFWHMInVertices**: Estimated Full Width at Half Maximum of the noise distribution in world vertices.	nidm:MaskMap	
* **spm:noiseFWHMInVoxels**: Estimated Full Width at Half Maximum of the noise distribution in voxels.	

I propose just inserting a "spatial smoothness" for clarity, and changing "distribution" -> "process", since smoothness/FWHM is really about the *spatial* properties of the process, and not the distribution of the process (at a single point, which is how we've mostly otherwise been using "distribution"):

* **spm:noiseFWHMInUnits**: Estimated Full Width at Half Maximum **spatial smoothness of the noise process** in world units.
* **spm:noiseFWHMInVertices**: Estimated Full Width at Half Maximum **spatial smoothness of the noise process** in world vertices.	
* **spm:noiseFWHMInVoxels**: Estimated Full Width at Half Maximum **spatial smoothness of the noise process** in voxels.	

**For FSL**, noise smoothness is never represented, rather only its inversely-related cousin "Determinant Lambda Hat" measure (in LaTeX, this is the essential RFT parameter $|\Lambda|$, its $|\hat{\Lambda}|$ estimator specifically).  For this I propose

* **fsl:noiseRoughnessInVoxels**:  Estimated spatial roughness of the noise process in voxel units, as measured by the determinant of the Lambda matrix (Lambda is the variance-covariance matrix of spatial derivatives of the noise process).

Is this too in the weeds?  It's accurate, though :)

Issue #128, opened 2014-07-24
Create a definition for "Resel"


cmaumet:
This discussion was initiated as a comment on the [spreadsheet of terms](https://docs.google.com/spreadsheets/d/16pC2cDsdxlzv2CzlNMtStqt5-xHwDEsU6MjZVxWhrE4/edit?usp=sharing).

---

From @nicholsn, `17:43 25 Apr`: 
Also, do we define "Resel" anywhere or how a Cluster is constructed? I assume this is using a threshold?

From @cmaumet, `09:53 8 May (edited 09:53 8 May)`: 
@nicholsn clusterSizeInxxx are attributes of ClusterLevelStatistic and there is therefore no definition of Cluster. Do you think this is something we should add? 

I agree we need a definition of "Resel" but I am not sure how this term would "link" with the rest of the vocabulary (other than being a child of prov:Entity). Do you have some ideas about this?

---

This thread is for the discussion on "resel" ("cluster is covered in #71).

nicholst:
Hi @khelm, @cmaumet - May I revive this issue and propose -- *if* it is deemed necessary -- to create a definition of resel.  Since we don't seem to have a definition of "voxel" I wonder if it is really needed; but since so many other terms refer to resel, it might just be worth nailing it down.  

* **nidm:resel**: "A virtual voxel having size equal to the Full Width at Half Maximum of the spatial smoothness of the noise process."

If we don't like "virtual voxel" then I guess we could use:

* **nidm:resel**: "A unit of space having size equal to the Full Width at Half Maximum of the spatial smoothness of the noise process."

This is inspired by other term curation efforts at #275.

PR #275, opened 2015-04-16
Definitions: RESEL-related and search volume terms

cmaumet:
Here are some proposals for some term curation:

* **fsl:dlh**:  As per discussions in #214, this term is replaced by **fsl:noiseRoughnessInVoxels**, having a definition of "Estimated spatial roughness of the noise process in voxel units, as measured by the determinant of the Lambda matrix in voxel units (Lambda is the variance-covariance matrix of spatial derivatives of the noise process)."

* **fsl:reselSizeInVoxels**: Propose "Size of one resel in units of voxels or verticies."
* **spm:reselSize**: Currently "Size of one resel in voxels or vertices.".  Also propose, for consistency, we re-name the term (possibly after giving it a numeric ID) of **spm:reselSizeInVoxels**. See discussion at #214 for discussion of resel's complementary measure, (noise) FWHM.

* **spm:searchVolumeInResels**: Propose "Size of search volume measured in resels."

Please review & comment on #128 for discussion on whether to create a resel term.

[Please review & comment on #269 for discussion on deletion of **fsl:searchVolumeInVoxels** & **spm:searchVolumeInVoxels**, to be replaced with **nidm:searchVolumeInVoxels**.] => merged


cmaumet:
Now implemented and ready to go. Thank you!

PR #305, opened 2015-04-16
Re-use "contrast weight matrix" from stato

cmaumet:
This is a proposal to re-use the term "contrast weight matrix" from STATO, in place of `nidm:ContrastWeights`. 

The STATO term was created following discussions at ISA-tools/stato#23 and is defined by "a contrast weight matrix is a information content entity which holds a set of contrast weight, coefficient used in a weighting sum of means defining a contrast".

How would you feel about this?


nicholst:
STATO's definition is awkward but accurate. +1 for re-use.

khelm:
"a contrast weight matrix is a information content entity which holds a set of contrast weight, coefficient used in a weighting sum of means defining a contrast".

Wow.  I think @nicholst calling it just "awkward" is rather charitable. I, on the other hand, would just say that it makes no sense as written and they should be alerted to fix it.

nicholst:
Well, the odd grammatical number issues were discussed but honestly I got lost and didn't follow up with the STATO folks.  See the discussion at [stato/#23](https://github.com/ISA-tools/stato/issues/23#issuecomment-65636555), where it's noted that all STATO class labels have to be singular.  

I've added a new comment to that [stato/#23](https://github.com/ISA-tools/stato/issues/23#issuecomment-94072259) and hopefully we can get it cleaned up.  It *is* the right concept, and it was originally added at our request, but between basing the definition on Wikipedia and shoe-horning it into the STATO world, it got mangled.

jbpoline:
I left a comment on the stato#23, but did not try to work on the "information content" thing - It does not make much sense to me though.

cmaumet:
@nicholst, @khelm, @jbpoline: I think it's important to note that the textual definition of **stato:'contrast weight matrix'** closely follows its logical definition:
```
stato:'contrast weight matrix' subClassOf obo:'has part' some stato:'contrast weight'
stato:'contrast weight matrix' subClassOf obo:'is about' stato:contrast
stato:'contrast weight matrix' subClassOf obo:'information content entity'
```

Where each stato (and obo) term is separately defined:
 - *stato:'contrast weight'*: "a contrast weight is a coefficient which multiplies a group mean, part of a linear combinaison defining a constrast as a weighted sum of group means, giving a 'weight' to a specific group mean hence the name."
 - *stato:contrast*: "A contrast is the weighted sum of group means, the c_j coefficients represent the assigned weights of the means (these must sum to 0 for orthogonal contrasts)"

cmaumet:
(Rebased on master).

I am going to merge this update as I think discussions regarding the definition will be carried on at ISA-tools/stato#23.

PR #306, opened 2015-04-16
Re-use error covariance structures from STATO

cmaumet:
Following discussions at ISA-tools/stato#28, this is a proposal to:
 - re-use **stato: 'covariance structure'** in place of **nidm:ErrorDependence**.
 - re-use **stato: 'compound symmetry covariance structure'** in place of **nidm:CompoundSymmetricNoise**.

@nicholst: could you help me in finding out if there is a match for some of these nidm terms:
image
in the current set of covariance structures defined in STATO:
image

(Once the discussion will be fully implemented at ISA-tools/stato#28, we will replace all nidm terms currently defined as children of *stato:'covariance structure'*.)


nicholst:
@cmaumet, surprisingly, these don't map so well.
- `nidm:ArbitrarilyCorrelatedError`: amazingly, STATO seems to have omitted SAS's "Unstructured" covariance structure.  I've raised this with an addition to STATO's ISA-tools/stat#28.  So, no reference for now.
- `nidm:ExchangeableError`: This can reference STATO's "compound symmetry", but in fact is more general -- compound symmetry is just about correlations/covariances, while "exchangeability" refers to the entire multivariate distribution.  "Gaussian variables that are exchangeable are compound symmetric."
- `nidm:IndependentError` - Again, no love in STATO; raised in my  ISA-tools/stat#28 update.
- `nidm:SeriallyCorrelatedError` - Finally!  This maps exactly to 'Toeplitz covariance structure'

jbpoline:
I was a little puzzled as well. 

+1 For SeriallyCorrelatedError == Toeplitz. (in case of stationarity, always assumed for fMRI...)

cmaumet:
Thank you @nicholst and @jbpoline! Re-use of `stato:'Toeplitz covariance structure'` is now implemented. #312 is open to remind us to replace the remaining error dependence terms once ISA-tools/stato#28 is closed.

I will now merge those changes.

PR #311, opened 2015-04-17
minor fixups - please review

yarikoptic:
I felt also like it might be a good time to start thinking about organizing existing .py helpers within nidm Python package/module


yarikoptic:
serialisation vs serialization was chosen as a dominant choice -- viva UK

PR #310, opened 2015-04-20
SPM-specific definitions (inc. "smallest significant cluster size in voxels")

cmaumet:
The following SPM-specific definitions are currently pending final vetting and I would say are ready for release:

 - **spm:searchVolumeReselsGeometry**: Description of geometry of search volume. As per [Worsley et al.](http://www.ncbi.nlm.nih.gov/pubmed/20408186), the first element is the Euler Characteristic of the search volume, the second element is twice the average caliper diameter, the third element is half the surface area, and the fourth element is the volume. With the exception of the first element (which is a unitless integer) all quantities are in units of Resels
 - **spm:smallestSignificantClusterSizeInVerticesFDR05**: Smallest cluster size in vertices that are significant at a false discovery rate corrected alpha value of 0.05.
 - **spm:smallestSignificantClusterSizeInVerticesFWE05**: Smallest cluster size in vertices significant at family-wise error rate corrected alpha value of 0.05
 - **spm:smallestSignificantClusterSizeInVoxelsFDR05**: Smallest cluster size in voxels significant at false discovery rate corrected alpha value of 0.05.
 - **spm:smallestSignificantClusterSizeInVoxelsFWE05**: Smallest cluster size in voxels significant at family-wise error corrected alpha value of 0.05.

Any +1 or comments on those definitions?


cmaumet:
Thank you! This is now implemented and ready for release.

PR #309, opened 2015-04-20
Definition of 'expected Number Of Clusters' (pending final vetting)

cmaumet:
I would just suggest a small edit in this definition (currently pending final vetting):

 - **nidm:'expected Number Of Clusters'**: Expected number of clusters in an excursion set **according to Random Field Theory**.

How do you feel about this?


cmaumet:
@nicholst and @jbpoline: thank you for your comments! I have implemented @nicholst's suggestion and set the status of those terms to "pending final vetting". Issue #314 is open to discuss how to best handle the reference to RFT (separate term or sub-case in the textual definition).

I will now merge those changes.

PR #308, opened 2015-04-20
Definition of 'cluster Size In Voxels' (pending final vetting)

cmaumet:
Another definition, pending final vetting, that I think is good to go:
  - **nidm:'cluster Size In Voxels'**: Number of voxels that make up the cluster.

What do you think?


cmaumet:
@khelm: As the "cluster connectivity criteria" is mentioned in the definition of `nidm:'Significant Cluster'` (of which 'cluster Size In Voxels' is an attribute), I think that we do not need to re-mention it in this definition. I am going to merge but if you think this should be further discussed please feel free to open a separate issue.

PR #307, opened 2015-04-20
Definition of "Resels Per Voxel Map" (metadata complete)

cmaumet:
This issue is open to current the definition of *nidm:'Resels Per Voxel Map'*. I would just like to include a small edit to comply with other map definitions:
 - **nidm:'Resels Per Voxel Map'**: A map whose value at each **element (e.g., pixel, voxel, vertex, or face)** ~~location~~ is the number of resels per voxel.

Otherwise, I think this is ready for release?


cmaumet:
Thank you! This is ready for release.

PR #173, opened 2015-04-20
noiseFWHM: see also "Full width at half maximum"

cmaumet:
**Term**: `noiseFWHM `
**Current definition**: "Estimated Full Width at Half Maximum of the noise distribution."
**See also**: [Full_Width_at_Half_Maximum](http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl#Full_Width_at_Half_Maximum)

---

@Jessica Turner `03:44 5 Mar 2013`
[Commenting on "see also: http://ncicb.nci.nih.gov/xml/owl/EVS/Thesaurus.owl#Full_Width_at_Half_Maximum"]

this is the definition of FWHM, not the FWHM of the noise, per se
	
@cmaumet `08:41 24 Apr`
Would it be better to have the NCICB FWHM term as a parent of noiseFWHM?
	
@nicholsn `22:02 24 Apr`
noiseFWHM seems more like an attribute of some entity whose FWHM was calculated. Child terms should stand in an "is-a" relationship with parent terms. Following this which would you choose?

noiseFWHM is-a noiseMeasure
or
noiseFWHM is-a FWHM
	
@cmaumet `11:30 25 Apr`
noiseFWHMInUnits is an attribute of MaskMap (but as commented elsewhere we probably want another type for that entity). 

I would say noiseFWHM is-a FWHM of the noise. Does that make sense?

### DIFF
--- a/nidm/nidm-results/terms/README.md
+++ b/nidm/nidm-results/terms/README.md
@@ -221,13 +221,6 @@ Thank you in advance for taking part in NIDM-Results term curation!
 </tr>
 <tr>
     <td><img src="../../../doc/content/specs/img/orange.png?raw=true"/>  </td>
-    <td>Under discussion at <a href="https://github.com/incf-nidash/nidm/pull/173">#173</a><br/><a href="https://github.com/incf-nidash/nidm//issues?&q='noise FWHM'"> [more] </a></td>
-    <td><b>nidm:'noise FWHM': </b>Estimated Full Width at Half Maximum of the noise distribution</td>
-    <td></td>
-    <td></td>
-</tr>
-<tr>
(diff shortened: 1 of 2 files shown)


cmaumet:
To be consistent with other nidm-results terms (e.g. `search Volume in ***`, `expected Number of *** per Cluster`...), this pull request delete the (unused) `nidm:noiseFWHM` parent (and only keeps the children: `spm:noiseFWHMInVoxels`, `spm:noiseFWHMInUnits`, `spm:NoiseFWHMInVertices`). 

I am going to merge now. If you think this should be further discussed, please feel free to open a new issue.

PR #145, opened 2015-04-20
Definition of  "coordinate1": use a generic term instead of "voxel"?

cmaumet:
**Term**: `coordinate1`
**Current definition**: "Coordinate along the first dimension in voxel units."
**Original comment**: [link](https://docs.google.com/spreadsheets/d/16pC2cDsdxlzv2CzlNMtStqt5-xHwDEsU6MjZVxWhrE4/edit?disco=AAAAAJVWUZ4)

---

@cmaumet `10:00 9 Apr`
Should we use "MapSubElement" in this definition instead of "voxels" to be more generic (cover the case of surfacic data in which we should talk about "vertices")?
	
@nicholsn  `20:23 18 Apr`
possibly, but would "MapSubElement" be a parent or child of "Coordinate"? 

[more comments on naming of "Map" not reported here]


cmaumet:
I have just implemented a new definition for "coordinate" (which replaced "coordinate1" since #270) as follows:
 - *nidm:coordinate*: A vector with one number per dimension. The first element corresponds to the coordinate along the first dimension measured in map elements (e.g., pixels, voxels, vertices, or faces), the second element to the coordinate along the second dimension etc.

This term is marked as "pending final vetting" and will be further discussed in a separate issue. Furthermore issue #322 is open to discuss the possibility to have "voxel"/"pixel"/"vertex" as an attribute in the model.

khelm:
I think what you are describing is a "Coordinate Vector", rather than a single coordinate, right?  A coordinate is the value of a point along a specified axis.

cmaumet:
Thanks @khelm! I kept the status of this term to "pending final vetting" (as the definition might need to be further reviewed) but upadted the label from "coordinate" to "coordinate Vector" in #324 for clarity.

PR #294, opened 2015-04-21
Definition of "k-conjunction inference" terms

cmaumet:
This issue is open to discuss the terms used for k-conjunction inference.

###### Proposal
Similarly to the definition of *nidm:ConjunctionInference* (cf. #134), what do you think of:
 - **spm:KConjunctionInference**: the process of testing for the joint significance of multiple effects, with emphasis on rejecting one or more (instead of all) of the respective null hypotheses as defined by the global null degree attribute. This process is also known as "global null conjunction" in SPM.
 - **nidm:globalNullDegree**: number of null hypothesis to be rejected in a k-conjunction inference.

?

I am happy to keep *KConjunctionInference* in the SPM namespace as I believe this type of conjunction is specific to SPM (but no strong feeling on this). But I think that if we keep *KConjunctionInference* in the `spm` namespace, then it would make more sense to also have *globalNullDegree* in the `spm` namespace.

###### Current definitions (for reference)
 - *spm:KConjunctionInference*: Inference testing for the joint significance of a subset of the effects.
 - *nidm:globalNullDegree*: \<undefined\>

### DIFF
--- a/nidm/nidm-results/terms/README.md
+++ b/nidm/nidm-results/terms/README.md
@@ -20,11 +20,6 @@ Thank you in advance for taking part in NIDM-Results term curation!
     <td><b>nidm:'Coordinate Space': </b>An entity with spatial attributes (e.g., dimensions, units, and voxel-to-world mapping) that provides context to a Map (e.g., a Statistic Map, a Contrast Map...)</td>
 </tr>
 <tr>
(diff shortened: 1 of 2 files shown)


cmaumet:
Thank you @khelm! I have just set the status to "ready for release" and will include it in #324 as part of NIDM-Results 1.0.0.

Issue #319, opened 2015-04-20
Harmonise FSL and SPM noise FWHM terms


cmaumet:
This issue is open to keep a track of discussions started at #214 regarding the possibility to harmonise the noise FWHM terms (and export) for SPM and FSL.

PR #301, opened 2015-04-15
fsl, spm and nidm namespaces

cmaumet:
Recently we have been discussing the appropriateness of having a given term in the SPM of FSL namespaces versus in the generic NIDM namespace. 

In this issue are listed all the remaining software-specific terms:
 - The first list "to keep in a software-specific namespace" contains the terms that I think should remain software-specific as they are only defined in the context of a given software and do not represent a general concept. 
 - In the second list "Only reported in one software but could be in NIDM namespace?" are the terms that we might consider general-enough to move them to the the nidm namespace.

Could you please have a look and let us know if you think some of these should be in the nidm namespace rather than being software-specific?

To keep in a software-specific namespace:
 - `spm:searchVolumeReselsGeometry` [[definition]](https://github.com/incf-nidash/nidm/blob/84b704507aa20dc9f46fe9df8b118acbeb35d9da/nidm/nidm-results/terms/nidm-results.owl#L1375)
 - `spm:GammaDifferenceHRF` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `spm:TemporalDerivative` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `spm:DispersionDerivative` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `fsl:GammaDifferenceHRF` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `fsl:TemporalDerivative` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `afni:GammaHRF` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `afni:BLOCK` [[definition]](https://github.com/incf-nidash/nidm/issues/281#issue-65667736)
 - `spm:KConjunctionInference` [[definition]](https://github.com/incf-nidash/nidm/issues/294#issuecomment-91461441)
 - `fsl:GaussianRunningLineDriftModel` [[definition]](https://github.com/incf-nidash/nidm/issues/298)
 - `fsl:featVersion` [[definition]](https://github.com/incf-nidash/nidm/issues/291)

Only reported in one software but could be in NIDM namespace? 
 - `spm:expectedNumberOfClusters`
 - `spm:expectedNumberOfVerticesPerCluster`
 - `spm:expectedNumberOfVoxelsPerCluster`
 - `spm:heightCriticalThresholdFDR05`
[...]

PR #328, opened 2015-05-11
Noise FWHM in nidm namespace

cmaumet:
This is a proposal to move the "noiseFWHM" terms (i.e. `spm:'noise FWHM in units'`, `spm:'noise FWHM in voxels'` and `spm:'noise FWHM in vertices'`) to the nidm namespace.

We initially created those terms in the SPM namespace because those values were only reported for SPM results. But:
 - Since #301, we have considered that terms representing generic concepts would be defined in the nidm namespace. 
 - Moreover, issue #319 and #214 (cf. in particular https://github.com/incf-nidash/nidm/pull/214#issuecomment-6179982) have proposed to also report `'noise FWHM in units'` and `'noise FWHM in voxels'` for FSL (implemented in https://github.com/cmaumet/nidm-results_fsl/commit/82c51f9625bf10c31ed894ff373bf91d9fd714e0).


nicholst:
+1

khelm:
Do the terms have the same definition in SPM and FSL?

cmaumet:
Yes, current definitions (for SPM) are generic and could also be used for FSL:
 - [spm:'noise FWHM In Voxels'](http://nidm.nidash.org/specs/nidm-results_100.html#dfn-spm_0000009): "Estimated Full Width at Half Maximum of the spatial smoothness of the noise process in voxels"
 - [spm:'noise FWHM In Units'](http://nidm.nidash.org/specs/nidm-results_100.html#dfn-spm_0000007): "Estimated Full Width at Half Maximum of the spatial smoothness of the noise process in world units"

khelm:
OK, thanks.  +1

nicholst:
For the record, FSL and SPM use different methods to estimate the FWHM (FSL uses correlation of residuals at pairs of adjacent voxels; SPM uses variance of the difference of residuals at pairs of adjacent voxels), but they are both trying to estimate the same intrinsic thing.

khelm:
Ok, so in a sense they have different "definitions".  This seems to be a case where we should decide an overall strategy for this issue.  My view is that this is the kind of info that one would want to capture in the provenance, especially since we (i.e., you) actually know the difference.  So in the case where we could write a generic definition to cover all cases, but the underlying method is different, could we use the generic definition language, but then have a second sentence saying what method is used?

cmaumet:
We discussed this on NIDASH call on May 11th.

cmaumet:
(Rebased on master). I am going to merge this as agreed on NIDASH call on May 11th.

PR #330, opened 2015-05-14
NIDM specifications: remove dependency on w3c website

cmaumet:
Recently, we (with @nicholst) experienced some delays when accessing the specifications at http://nidm.nidash.org/specs. It turned out that these resulted from difficulties in accessing the `respec-w3c-common` file on the w3c website `http://www.w3.org/Tools/respec/`. To avoid this issue in the future, this pull request updates the specifications so that they now rely on a "local" copy of `respec-w3c-common`.


nicholst:
+1

nicholsn:
​good idea, +1​

khelm:
+1

cmaumet:
Thank you!

PR #331, opened 2015-05-18
NIDM-Results tests: compare against a given version of nidm-results

cmaumet:
Now that we are dealing with several versions of NIDM-Results, this PR updates the test procedure so that it automatically checks which version is in use in the test data (looking at the `softwareVersion` attribute in the bundle) and compare with the corresponding tag (e.g. for "1.0.0", compare with https://github.com/incf-nidash/nidm/tree/NIDM-Results_1.0.0).


cmaumet:
I am going to merge this now so that https://github.com/incf-nidash/nidm-results_fsl and https://github.com/incf-nidash/nidm-results_spm can work with the main nidm repository (rather than the `test_based_on_nidm_version` branch).

PR #347, opened 2015-10-20
Fixes in SPM examples: equivalent threshold and reconciliation of beta maps

cmaumet:
This pull request fixes a set of issues identified with @gllmflndn when testing the SPM exporter with UUIDs (cf. https://github.com/incf-nidash/nidm-results_spm/pull/20):
 - Reconciliation of beta maps was failing
 - equivalent thresholds (both height and extent) where not linked properly to the main threshold
 - [minor] SPM software version updated to "12.12.1"


cmaumet:
I am going to merge these fixes so that we can carry on testing at https://github.com/incf-nidash/nidm-results_spm/pull/20.

PR #344, opened 2015-10-15
Add links to comment the specification document

cmaumet:
As discussed during [NIDASH Task Force meeting](https://docs.google.com/document/d/1YWUd1Uupjf1ZBX0Nz4MeC1XAYcKdrICvLtO2y1sVdao/edit?usp=sharing), this pull request adds two links after each definition in the development version of the NIDM-Results spec:
 - a link (star) to look for issues discussing this term 
 - a link (plus sign) to create a new issue to discuss this term

Example:
screen shot 2015-10-15 at 15 13 16

PR #348, opened 2015-10-29
NIDM-Results specs: updated charts, various formatting fixes

cmaumet:
This PR implements update of the NIDM-Results specification as suggested by @gllmflndn, including updates in:
 - Part of the introduction
 - Format the definitions (new line before stating prov:type)
 - Simplification of the main UML chart
 - Order of the concepts in the tables
 - Add an example of first-level design matrix
 - Remove broken links
 
Additionally the following updates have also been implemented:
 - List children and parent for each class
 - A new chart per section listing all optional entities (removed from main graph)


cmaumet:
Preview: [NIDM-Results updated spec](https://htmlpreview.github.io/?https://github.com/cmaumet/nidm/blob/spec_comments_gllmflndn/doc/content/specs/nidm-results_dev.html)

nicholst:
Trivial tweaks to possibly fix later:
* The side bar still reads "NI-DM Working Draft" instead of NIDM (no dash)
* There is no header... isn't there usually some sort of title or something, or an icon?  When I view it, it simply starts "In neuroimaging, the standard analysis ..." no title or anything.

cmaumet:
Thank you @nicholst! I think that the header issue is related to the preview. When loading locally the usual headings are displayed properly.

cmaumet:
To replace "NI-DM" by "NIDM" in the sidebar, we would need someone with photoshop to edit `doc/content/specs/stylesheet/logo-NIDM-WD.psd` and export to `doc/content/specs/stylesheet/logo-NIDM-WD.png`. 

Unfortunately gimp does not seem to recognise text layers created with photoshop...

cmaumet:
Mention of "NI-DM" in the sidebar is now fixed: thank you @nicholst!

cmaumet:
Added:
 - `nidm:Map` and `nidm:Map Header` examples
 - A title for each example

Fixed:
 - Missing `prov:atLocation` attribute in `nidm:Map`

cmaumet:
I will go ahead and merge this. (We can continue updating the specs with a new PR.)

cmaumet:
The updated spec is now available at: http://nidm.nidash.org/specs/nidm-results_dev.html

cmaumet:
As a side note: we can use rawgit for the previews to avoid the issue of javascript not loading with htmlpreview. For this PR we would have used the following URL: https://rawgit.com/cmaumet/nidm/78e3ccda21f1722f3979193090e1ce004f09f325/doc/content/specs/nidm-results_dev.html

nicholst:
Super!

I like the 'star' links especially, to each term as it appears. BTW, what
is the '+' there for?  Does it always follow the star?

On Fri, Oct 30, 2015 at 11:13 AM, Camille Maumet <notifications@github.com>
wrote:

> The updated spec is now available at:
> http://nidm.nidash.org/specs/nidm-results_dev.html
>
> —
> <https://github.com/incf-nidash/nidm/pull/348#issuecomment-152498263>.
>



-- 
Thomas Nichols, PhD

Fx,  +44 24 7652 4532

nicholst:
Not sure if this is the right place to post this, but found an error in the definition of `nidm:'Contrast Explained Mean Square Map' `

It says it is 

> A map whose value at each element (e.g., pixel, voxel, vertex, or face) is the extra sum of squares divided by the effect degrees of freedom (i.e. the denumerator of an F-statistic).

which is correct *except* that it is the _numerator_ and not denominator (which anyway is misspelled).

cmaumet:
@nicholst: The 'star' and the '+' links were added in https://github.com/incf-nidash/nidm/pull/344 to help with getting feedback on the specs following discussion at the [Oct 2015 NIDASH Task Force meeting](https://docs.google.com/document/d/1YWUd1Uupjf1ZBX0Nz4MeC1XAYcKdrICvLtO2y1sVdao/edit?usp=sharing):
 - 'star' links to the issues currently open to discuss this term 
 - '+' creates a new issue
Each term has a 'star' and a '+' when first defined.

cmaumet:
@nicholst: Thank you for catching that issue in the definition of 'Contrast Explained Mean Square Map'. https://github.com/incf-nidash/nidm/pull/349 implements the suggested fix.

proccaserra:
@cmaumet @nicholst : in section of 3.2.4.1 nidm:'Error Distribution' in the specification document https://rawgit.com/cmaumet/nidm/78e3ccda21f1722f3979193090e1ce004f09f325/doc/content/specs/nidm-results_dev.html, I was wondering if declaring a new relation 'follows_probability_distribution' between 'error distribution' and STATO probability distribution would be a simpler way to represent this.
It has potentially several advantages:
i. avoid multiple parenthood caused by is_a relation that is currently used,
ii. allows more straightforward reuse of 'STATO' terms, directly importing the terms.
iii. potential gains in 'maintainability'
iv. easier integration with other semweb datasets.

all the best,
Philippe

nicholst:
@proccaserra I'm ignorant of the ontological issues, but statistically:  A random variable follows a distribution, a distribution doesn't follow a distribution.  I like the idea of a  'follows_probability_distribution' relation, but then it would need to be like 'model error'  'follows_probability_distribution' <a STATO Distribution>.

@cmaumet - Can you remember why exactly why we cloned and just referenced STATO's distributions, and didn't directly use them?

proccaserra:
@nicholst, indeed. the relation should be between a random variable and a probability distribution.

cmaumet:
@proccaserra: thank you for catching this. @nicholst: I cannot recall why we did not directly use the STATO classes here and don't see why we should not. The only missing bit in stato is the nidm:'Non Parametric Symmetric Distribution' that will be required for NIDM-Results non-parametric extension.

I have just opened #350 with an update proposal: please let me know what you think by commenting over there. Thank you!

khelm:
We didn't directly use the STATO terms because they use numerical ID's rather than human readable ones.

nicholst:
Ah! Well that's no longer an issue...

PR #349, opened 2015-10-30
Definition of 'Contrast Explained Mean Square Map': numerator

cmaumet:
This PR fixes the definition of nidm:'Contrast Explained Mean Square Map' as suggested by @nicholst in https://github.com/incf-nidash/nidm/pull/348#issuecomment-152526658 ("denominator" replaced by "numerator").


nicholst:
+1 to merge

PR #262, opened 2015-03-05
Re-use STATO "statistic"

cmaumet:
This pull request implements the re-use of STATO `statistic` terms as a value for `nidm:statisticType` object property used in the `nidm:StatisticMap` entity ([current implementation](http://nidm.nidash.org/specs/nidm-results_dev.html#dfn-nidm-statisticmap), [proposed update preview](http://htmlpreview.github.io/?https://raw.githubusercontent.com/cmaumet/nidm/blob/reuse_stato_statistic/doc/content/specs/nidm-results_dev.html#dfn-nidm-statisticmap)).

PR #350, opened 2015-10-30
Re-use STATO probability distribution terms for Error Distribution

cmaumet:
Following discussion at https://github.com/incf-nidash/nidm/pull/348#issuecomment-152537679 with @proccaserra and @nicholst, this PR implements the re-use of STATO `probability distribution` terms as a value for `nidm:'has Error Distribution'` object property used in the `nidm:'Error Model'` entity.


cmaumet:
This is similar to what was done at #262 (to re-use STATO statistic terms), #252 (to re-use STATO estimation methods)...

khelm:
I agree that we should re-use STATO terms as much as possible, but I just wish they were edited a bit better.  I've seen a number of instances like this:

"definition "A probability distribution is a information content entity specifies the probability of the value of a random variable."

It's not that it's wrong in content, but it's not grammatically correct.

nicholst:
It is unfortunate. But my impression is that there is exactly one person (a tiny FTE) that works on it. Suggest an edit at https://github.com/ISA-tools/stato in my experience they (the one guy) is very responsive. 

-Tom

proccaserra:
:) we will fix this as shortly and push it. We are in fact 2, Alejandra and myself, with input from Orlaith and of course you!
Another comment: it has to do with 'fMRI Design Type' and the notion of 'block based design'.
it seems to be related to the delivery of a treatment/stimulus (single exposure or repeated exposure).
I was wondering if the current term is meant to cover the declaration of 'blocking variable' and is related to 'randomized block design' ?

cmaumet:
This was discussed on NIDM call on Nov 2nd and was agreed upon.

PR #360, opened 2015-12-01
Fix: truncated sha sum in examples

cmaumet:
This PR removes `...` from sha-sum values in the examples. I can't recall when those sha-sum were truncated but in practice, they should not contain triple-dots.


cmaumet:
(Rebased on master)

cmaumet:
(Rebased on master).

Issue #353, opened 2015-11-03
http://incf.org/ns/nidash/nidm resolves to nonexistent http://ns.nidash.org


nicholst:
For no particular reason, I tried resolving the URI of a definition, e.g. http://purl.org/nidash/nidm#NIDM_0000127 (nidm_version).  This resolved to 

> https://raw.githubusercontent.com/incf-nidash/nidm/master/nidm/nidm.owl#NIDM_0000127

and that lists this as the ontology: 

> http://incf.org/ns/nidash/nidm

and _that_ resolves to

> http://ns.nidash.org

which is "404 - There isn't a GitHub Pages site here."

@nicholsn - I know it doesn't matter officially, but can http://ns.nidash.org be created and offer something useful (nidm.owl?)  Or should we change the redirect from INCF?

@jbpoline - As a side note, will the reorganization of the INCF funding, will "NIDASH" still exist as an entity?

satra:
@nicholst - we just need to setup a static site for ns.nidash.org (on github).

satra:
@nicholst - more importantly we need to decide where the purl resolves and how.

also we need to be able to provide the human readable version when asked by a human and the machine readable version when ask by a machine.

satra:
also https://github.com/incf-nidash/nidm/blob/master/nidm/nidm.owl has not been updated.

nicholsn:
Also the purl.org admin interface is currently down so I can't update the
purl 303 redirect but will once it is back up. See:
https://groups.google.com/forum/m/#!topic/persistenturls/ZPE8_udZwnY

Will also send a PR to update our top level owl file with more meaningful
info and imports - stll need to handle the issues satra mentions though...

sent via mobile, please excuse brevity and typos

also https://github.com/incf-nidash/nidm/blob/master/nidm/nidm.owl has not
been updated.

—
Reply to this email directly or view it on GitHub
<https://github.com/incf-nidash/nidm/issues/353#issuecomment-153374718>.

jbpoline:
@nicholst : yes nidash will survive (unless we want it to stop).  
@nicholsn @satra @nicholst : I had done the experiment of trying to find for the relation between a term and it's definition and its unique identifier, which was not straightforward, and I had in mind for the nidm sprint to do the documentation / tool to make the link very easy. We should also keep in mind the larger nidm-terms project for this.

khelm:
[...]

PR #363, opened 2015-12-09
Import nidm-results and nidm-experiment in nidm.owl

cmaumet:
As discussed in #353, on the NIDM calls and by emails with @nicholsn and @khelm, this PR:
 - import `nidm-results.owl` and `nidm-experiment.owl` in `nidm.owl` 

This is one step forward in solving #353. But, it does not yet solve the issue as, as far as I know, imports are only useful when loading the owl file in Protege. 

For instance if we look-up http://incf.org/ns/nidash/nidm#NIDM_0000127, this will still not resolve to the term definition (as the term definition is "hidden" behind the nidm-results.owl import). 

One solution would be to create released versions of `nidm.owl` in which we actually include all the released terms.... @nicholsn: can you think of another solution? Thanks!


satra:
just leaving this here for a read: http://renaud.delbru.fr/doc/pub/rw-2013.pdf

nicholsn:
@cmaumet, I like the idea of creating a "materialized" release that includes all the imported terms, etc.

At least for now, I think it would make sense just to include NIDM Results until we firm up other terms.

nicholsn:
Also, in terms of a release we still do not have access to the admin side purl.org so we can't change where the nidm namespace IRI points

PR #365, opened 2015-12-09
Highlight terms mentioned as part of the definitions

AlexBowring:
The document has been updated so that whenever a term is used that is defined in the NIDM-model it links to the definition.


cmaumet:
For example, definition of `nidm:'has Drift Model'` is now:
 - "A property that associates a **'Drift Model'** to a **'Design Matrix'** (only used for first-level fMRI experiments)."

instead of:
 - "A property that associates a drift model to a design matrix (only used for first-level fMRI experiments)."

cmaumet:
(Replaced by #366).

PR #366, opened 2015-12-14
Highlight terms mentioned as part of the definitions

cmaumet:
This  PR replaces #365 initiated by @AlexBowring. The same code is included but rebased on master.


cmaumet:
@AlexBowring is also added to the contributor list for NIDM-Results.

cmaumet:
Example of update:
screen shot 2015-12-14 at 12 01 07


replaces:
screen shot 2015-12-14 at 12 00 19

cmaumet:
Thanks for the edits @AlexBowring! Let's merge.

PR #395, opened 2016-09-02
Fix install error

cmaumet:
As in https://github.com/incf-nidash/nidmresults-spm/pull/33, this pull request add an update of the setuptools package before install of the requirements to avoid the `html5lib requires setuptools version 18.5 or above; please upgrade before installing (you have 12.0.5)` error.


cmaumet:
I am going to merge this small fix that makes the test runnable again.

PR #396, opened 2016-09-02
added qibo and sio import files and updated nidm-exp owl file

khelm:
Added draft versions of the qibo and sio import files and updated the nidm-experimental.owl file to correspond to the data model lucid chart ver6.


cmaumet:
@khelm: following our discussions by email, I had a look at the `nidm-experiment.owl` file and found a couple of typos (that were causing errors when trying to open the file in Protege). 

Can you see a checkbox on the left of this PR saying `Allow edits from maintainers.`? If so (this is a new feature so not sure if it's available here yet), could you enable this so that I can contribute my edits directly in this PR?

cmaumet:
I have just found this validator that is quite useful to track down this type of issues: http://mowl-power.cs.man.ac.uk:8080/validator/ (with more precise error messages than what protege typically return).

khelm:
Thanks very much, I've checked the "Allow edits..." box.
How does that validator handle imports?  Does it just ignore them?
Also, when I'm working on this locally it would seem to be easiest to have the imports point to local files rather than the URL's since I'm also editing the imports files as I go.  Is this what you do or is there an easier way?

cmaumet:
Thank you for allowing the edits! I have just pushed my small fixes (cf. ef250c7). 

I think the validator is just checking the overall syntax (e.g. missing `;`) but not actually looking inside the import files.

The local path to the import files can be defined in https://github.com/incf-nidash/nidm/blob/master/nidm/nidm-experiment/terms/catalog-v001.xml so that Protege loads them directly when opening the owl file. Is that what you were looking for? Can you point me to an example of import using a URL?

khelm:
I saw a section at the top of the nidm-results.owl file after the prefixes and base were defined there was a block of code that defined the imports and these were all purl links.  I'll look in the catalog file you mentioned.

khelm:
I just looked in the catalog-v001.xml file and I see now - the purl uri are mapped to local uri's.  Got it, thanks.

cmaumet:
I understand now! Yes, that's how we use the catalog for NIDM-Results.

khelm:
Also I want to go through all of the imports to make sure they make sense for nidm-experiment.  Are these just the ones used for Results or are they are first draft for Experiment?

cmaumet:
@khelm: Most of the import files in the `import` folder are only referenced in the NIDM-Results owl file. The current NIDM-experiment owl file, imports IAO terms (to record the curation status) and PROV (cf. https://github.com/incf-nidash/nidm/blob/master/nidm/nidm-experiment/terms/nidm-experiment.owl#L35-L39). 

If that makes things easier, each component could also have its own `import` folder.

khelm:
I think it would be a good idea for each to have their own imports folder.  That would get around the issue of having to deal with un-needed terms from other components.

cmaumet:
@khelm: I have just made a couple of commits to update the `nidmresults` library and fix compatibility with python 2.

One of the test in this PR fails because the nidm-experiment HTML spec is out to date compared to your owl file. You can automatically update the HTML spec by running:
```
python nidm/nidm-experiment/scripts/create_expe_specification.py
```

Would you like to try this on your machine to make sure everything works fine? And if it does you can then push the new HTML document for everyone to look at.

khelm:
Ok, thanks for the heads up.  I'll run that locally this morning and let you know what happens.

khelm:
This throws an error, but I think it's a simple fix.  Here's the error:
karl@karl-Satellite-C655D:~/Work/INCF/nidm/nidm/nidm/nidm-experiment/scripts$ python create_expe_specification.py 
Traceback (most recent call last):
  File "create_expe_specification.py", line 22, in <module>
    from owl_to_webpage import OwlSpecification
  File "/home/karl/Work/INCF/nidm/nidm/scripts/owl_to_webpage.py", line 3, in <module>
    from nidmresults.owl.owl_reader import OwlReader
ImportError: No module named nidmresults.owl.owl_reader

khelm:
The version I have seems to be missing code to do the imports required in these two statements:
from nidmresults.owl.owl_reader import OwlReader
from nidmresults.objects.constants_rdflib import *
in .../nidm/scripts/owl_to_webpage.py

cmaumet:
The `nidmresults` module is available separatly, you can install it with pip:
```
pip install nidmresults
```

But because I have made a few updates (not released yet), I would suggest installing the dev version with:
```
pip install --upgrade --no-deps git+https://github.com/cmaumet/nidmresults.git@no_std_space_coord#egg=nidmresults 
```

khelm:
I installed all of the dependencies and the code now runs though it gives warnings for terms that I've replaced with other terms or removed for now because the structure was not ready for them yet.
 Examples:
4: UserWarning: No PROV type for class: nidm:InvestigationCollection
  self.graph.qname(class_name))
/home/karl/anaconda/lib/python2.7/site-packages/nidmresults/owl/owl_reader.py:21

And there are also a lot of PROV warnings:
4: UserWarning: No PROV type for class: nidm:DiffusionWeighted
  self.graph.qname(class_name))

What needs to be fixed?

cmaumet:
Hi @khelm: those warnings are fine, they are just raised because some of the terms we previously had in the specification are no longer available. I would suggest that we wait until the terms are stabilised a bit more before we decide how to split the specification into sub-sections and update this code. 

But those warnings should not prevent the tests from passing. Do they pass on your machine? If so, can you upload the updated HTML spec?

khelm:
Hi @cmaumet what file is this and where is it located?

cmaumet:
Hi @khelm, if the `create_expe_specification.py` script ran without error, you should see the updated HTML file with `git status`:
```
$ git status 
On branch add_nidm_experiment_terms_draft1
Your branch is up-to-date with 'origin/add_nidm_experiment_terms_draft1'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   doc/content/specs/nidm-experiment_dev.html

no changes added to commit (use "git add" and/or "git commit -a")
```

You can commit and push the `doc` folder in which the HTML specs are stored with:
```
$ git add doc/
$ git commit -m "python nidm/nidm-experiment/scripts/create_expe_specification.py"
$ git push origin add_nidm_experiment_terms_draft1
```

smpadhy:
Hi @khelm @cmaumet: I checked out this branch khelm:add_nidm_experiment_terms_draft1, when I opened the nidm/nidm-experiment/terms/nidm-experiment.owl file in protege, it was able to open the file with ontology. However, when I tried to open using 'open from URL' and specified the URL (https://raw.githubusercontent.com/khelm/nidm/add_nidm_experiment_terms_draft1/nidm/nidm-experiment/terms/nidm-experiment.owl), it showed me messages : "The system couldn't locate the ontology:http://purl.org/nidash/nidm/crypto_import.owl. 
I read previous comment of @cmaumet on resolving imports using category-001.xml. Is there a way to import the owl files not locally but from a URL?  In my app, I will be providing URL to the OWL file. So, if import URLs are not local would be useful.

cmaumet:
Hi @smpadhy!

Yes, the failing import is something we discussed with @khelm last week, he is planning to make another set of updates in this PR which will solve the issue.

Once the owl file is ready for release then all the terms will be copied to the main file so that imports are not required anymore (see for example the [NIDM-Results released owl file](https://github.com/incf-nidash/nidm/blob/master/nidm/nidm-results/terms/releases/nidm-results_130.owl)). So, the local imports are only used as we are developing the ontology. Does that make sense?

The validator is up again: http://mowl-power.cs.man.ac.uk:8080/validator/ :)

smpadhy:
@cmaumet Thank you for the details. The plan sounds great. I will wait for the final draft.

khelm:
@smpadhy the current version is now in decent shape. The assessment instrument section is going to be rearranged in the near future and the terms are still under review, but it makes sense to take a look at this version.

smpadhy:
@khelm @cmaumet Thank you for letting me know and the modifications. When I opened the nidm-experiment OWL file in Protege, I am seeing some error in the log such as "Illegal redeclarations of entities: resuse of entity http://ww.w2.org/ns/prov#wasRevisionOf" in punning not allowed [ObjectProperty: ...] ". However, the OWL got loaded to Protege and I can see the terms. I am not sure if that error has any effect on viewing the terms.

As I had mentioned in my earlier comments, protege still not able to resolve the imports when I open the file by URL. As I will be fetching the owl file from github in brainverse app, parseing it and using it to build a RDF graph, a single owl file would be helpful. As @cmaumet mentioned earlier, the plan is to generate a single owl file with all terms and no imports similar to nidm-results.owl. I will wait for the final draft. In the meantime, I will use the current file start exploring it.

khelm:
Ok, thanks for the heads up  - I haven't seen that error and I'll look into it.  Yes, this finally will be a single file as @cmaumet  mentioned.

PR #404, opened 2017-03-14
Crypto namespace ends with slash not hash

cmaumet:
This is an update suggested by @gllmflndn through the SPM NIDM exporter (cf. https://github.com/incf-nidash/nidmresults-spm/pull/42). 

NIDM-Results is currently using `http://id.loc.gov/vocabulary/preservation/cryptographicHashFunctions#` as base URI for the crypto namespace though, according to the [corresponding website](http://id.loc.gov/vocabulary/preservation/cryptographicHashFunctions/sha512), we should use `http://id.loc.gov/vocabulary/preservation/cryptographicHashFunctions/` (with a final `/`).

This PR updates the model and examples to use the official namespace. 

This change will break backward compatibility so should be included in the next major release of NIDM-Results. In practice, I don't think that any of our tools is using this attribute yet.


satra:
@cmaumet - perhaps we can do a few things:

1. make a copy of the external IRI info so that in the event they go away for whatever reason, we can run an update.
2. write a tool that checks external IRIs and compares them to local ones whenever we run our tests.
3. write an update tool that does some of these updates automatically for code and data.

as part of the plan for this year on repronim's agenda, we will be building a suite of nidm tools (validator, js library. py library, rest service). we probably won't get to all of that till we have @repronim/brainverse in a usable state, but once the base architecture and interface are ready we can start working on these tools (cc: @smpadhy).

cmaumet:
@satra, @smpadhy: Yes! We would definitely benefit from tools that will make our model more robust to dependencies on external vocabularies.

As this seems to be a middle-term goal, maybe we can merge this fix for now and keep your suggestions open as a new issue? 

By the way, is the repronim agenda publicly available? Sorry if I missed it... I am just thinking it would be great to have an overview so that we can better coordinate our efforts.

satra:
@cmaumet - re:fix yes please fix. 

re: agenda - it will come online here: https://github.com/ReproNim/TRD2/projects (this is specifically related to Project 2. ReproNim's overall architecture is here: repronim.org)

PR #416, opened 2017-07-07
Camel case preferred prefixes from external ontologies

cmaumet:
As discussed with @nicholst, @bthirion and @gllmflndn, this PR updates the preferred prefixes to be always camel cased (even if the original label was not using camel case, e.g. for imported terms such as OBO's). 

This update does not change the underlying model (and therefore does not break backward compatibility). The camel case prefixes will be used in the NIDM-Results minimal JSON API.


satra:
@cmaumet - should we make this a jsonld context file so that it can be reused immediately as context in other jsonld files?

cmaumet:
@satra: +1, now recorded in https://github.com/incf-nidash/nidm/issues/418.

cmaumet:
We discussed this PR in [NIDM call on July 17th](https://docs.google.com/document/d/1x2M_fMAfIkxJDP_RRhUuNNvymzke5f_lP7DIfQj2xWY/edit?usp=sharing).

PR #417, opened 2017-07-07
fix: some parsers don't like a # in @base

satra:


cmaumet:
Hi @satra! I remember encountering an error message like this in the past but I thought this was caused by a bug in early versions of rdflib (that was then solved). But are you saying here that # are indeed forbidden for @base?

satra:
@cmaumet - according to the spec it doesn't really seem so. i just checked again and indeed it is the same as prefix. but several things have raised a warning with the # in there (i wish the parsers would fix it). the other thing is that we don't use the base anywhere in that document. so perhaps we can just remove it for now.

cmaumet:
+1 to remove base as suggested. Let's add this to the next minor release of NIDM-Results?

satra:
i'm going to close this PR as this is not the solution. we can remove @base as the shorter term solution is implicit base is not being used.

PR #431, opened 2017-11-13
cleaned up import files and placed imports into NIDM graph. Cleaned u…

khelm:
Organized imported terms to fall under Prov-O graph. Cleaned up duplicate terms. Removed redundancies and unused terms.  Also am now treating prov-o as an import so that this can be worked on without access to online version (this can be reversed later).


khelm:
In response to the Travis errors, I've run the suggested code, but that didn't result in any file modifications.

Issue #461, opened 2018-04-23
Error when updating the website


cmaumet:
This issue was identified by @khelm. When updating the online version of the website, we get the following error:

```
$ fab serve
[localhost] local: rm -rf output
[localhost] local: mkdir output
[localhost] local: pelican content/ -s pelicanconf.py
WARNING: Feeds generated without SITEURL set properly may not be valid
ERROR: Skipping specs/README.md: could not find information about 'title'
WARNING: There are 1 items with empty slug: 
  | /Users/camaumet/Softs/nidm-specs/doc/content/pages/home.md
CRITICAL: UndefinedError: '_' is undefined

Fatal error: local() encountered an error (return code 1) while executing 'pelican content/ -s pelicanconf.py'
```

I've tried with both Python 2 and Python 3 and get the same error.

From what I can see this is a problem with the theme that we are using for the website (i.e. if I comment out `# THEME = 'pelican-bootstrap3` in pelicanconf.py the error goes away *but* we loose all the nice formatting...). 

Help would be greatly appreciated

khelm:
@cmaumet - Got rid of the error by:
1) git clone https://github.com/getpelican/pelican-plugins.git pelican-plugins
2) pip install jinja2 -U (to 2.10)
3) pip install fabric -U (to 2.1.3)
4) pip install python-gettext

added the following lines to pelicanconf.py
```
PLUGIN_PATHS = ['/path/to/pelican-plugins', ]
PLUGINS = ['i18n_subsites', ]
JINJA_ENVIRONMENT = {
    'extensions': ['jinja2.ext.i18n'],
}
```
A problem is that I couldn't use a relative path to point to the pelican-plugins
directory. We'll have to discuss how to deal with this.
I also got rid of the warnings by adding 
```
Title:
Date:
```
to /specs/README.md and 
```
Slug:
```
to .../nidm/nidm/doc/content/pages/home.md

In figuring this out, I found that updating fabric completely 
breaks fabfile.py as ver2.x is not backward compatible to
ver1.0, so I have also updated fabfile.py for fabric ver2.x.

cmaumet:
This issue was fixed by @khelm in https://github.com/incf-nidash/nidm-specs/pull/467. Thanks!

PR #467, opened 2018-05-29
fix webpage generation

khelm:
Updated packages and modified fabfile.py and pelicanconf.py files to reflect new package requirements so that web pages are now created.  See issue #461 for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+doc/pelican-plugins
+doc/pelican-themes
+
+
 *.tar.gz
 
 .DS_Store
--- a/doc/content/pages/home.md
+++ b/doc/content/pages/home.md
@@ -1,5 +1,6 @@
-Title:
+Title: NIDM
 URL:
+Slug: NIDM
 save_as: index.html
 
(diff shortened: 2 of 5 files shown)


cmaumet:
Hi @khelm, thanks for those updates! It's great to see the website code working again! 

I think I've found a solution to get rid of the full paths. As far as I understand, the issue was related to the fact that calls to `local` ignores the current conda environment. I've added optional arguments to define path to the virtual environment and the environment name:

Example of call if using a conda environment call `nidm2`:
```
fab -H 127.0.0.1 serve --ve=/Users/camaumet/anaconda2/envs/nidm2/bin/ --envname=nidm2
```

Without environment (i.e. when pelican is available by default):
```
fab -H 127.0.0.1 serve 
```

References (maybe useful later on): [prefix in fabric2](http://docs.pyinvoke.org/en/latest/api/context.html#invoke.context.Context.prefix), [Context manager in fabric 1](http://docs.fabfile.org/en/1.14/api/core/context_managers.html).

khelm:
Thanks @cmaumet for these updates!  I also tried `c.run`, and had no success - I think that is used to run commands on the remote server only, but I haven't looked into this extensively.

I'll have to play around with the python aspect a bit more as I'm not using virtual environments at the moment.  I should probably create one just for nidm as you have.  

In `800e002` I see that you created a relative path for the pelican-plugins directory.  So you are putting that directory into the nidm directory?

cmaumet:
Hi @khelm! Following your comment, I've moved the `pelican-plugins` folder under `doc` (and also added the folder to `.gitignore`). Do you think this is now ready for merge?

PR #468, opened 2018-05-29
new UpdateExpTermReadme.py file

khelm:
Created a modified UpdateTermReadme.py file for NIDM-Exp called UpdateExpTermReadme.py under the nidm-experiment path.
This will require changes to scripts/refresh.py, which I'm leaving for @cmaumet


khelm:
Yes, thanks.  I will follow up.  I wanted to make sure that `refresh.py` was calling `UpdateExpTermReadme.py` correctly before I spent time on that error.

cmaumet:
Great!! So happy to see the curation page implemented for NIDM-experiment: https://github.com/incf-nidash/nidm-specs/blob/master/nidm/nidm-experiment/terms/README.md.

khelm:
Yes, thanks for taking the time so we could sort out the pelican issue at the hackathon!

PR #469, opened 2018-05-31
Assess usage type

khelm:
Added Assessment Usage Type.  Got rid of individual terms for each type of Acquisition. Updated create_expe_specifications.py.


khelm:
This is the update that @dbkeator and @khelm worked on at the Montreal Hackathon.

PR #471, opened 2018-08-07
website: Set path to pelican as optional argument

cmaumet:
Fixed built with @khelm at the Montreal neuroinformatics hackathon. 

Fixes the webpage generation. 

This updates the arguments to the serve function to add an optional argument specifying the full path to the pelican command. The fix setting the conda environement using a full path (proposed in #467) was causing an internal conda error on @khelm's machine (KeyError 'PATH').

### DIFF
--- a/doc/fabfile.py
+++ b/doc/fabfile.py
@@ -58,10 +58,10 @@ def serve(c, fpp=''):
 
 
 @task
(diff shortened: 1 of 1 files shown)


cmaumet:
Tested with @khelm, let's merge!

PR #479, opened 2018-09-28
Add JSON-LD 1.1 exports for NIDM-Results examples

cmaumet:
This PR converts all our turtle examples into JSON-LD using an automatically generated context (from the OWL file) with type-indexing (JSON-LD 1.1 feature) to ease readability of the NIDM documents.


cmaumet:
@satra: Here is a complete demo. There is more cleanup to do, in particular with:
 - Boolean types (e.g. [here](https://github.com/incf-nidash/nidm-specs/blob/8b48784a26039216eac669d1539511e04cea1c46/nidm/nidm-results/spm/spm_results.json#L1468))
 - String containing JSON vectors (e.g. [here](https://github.com/incf-nidash/nidm-specs/blob/8b48784a26039216eac669d1539511e04cea1c46/nidm/nidm-results/spm/spm_results.json#L834)).
 - Object properties (e.g. [here](https://github.com/incf-nidash/nidm-specs/blob/8b48784a26039216eac669d1539511e04cea1c46/nidm/nidm-results/spm/spm_results.json#L874-L876))
 - more?

that are not condensed according to the context. Could this be an issue in the conversion ttl > JSON-LD via rdflib?

satra:
regarding `xsd:string` we should see why it's not behaving properly. for the spm case, i can see that the version is a number not a string, but in general we should force that to be a string, since we cannot guarantee appropriate versioning for all software.

cmaumet:
> @cmaumet - could we get the nidmr context creator merged in a separate PR, so that the jsonld documents created do not have to be embed the entire context?

For testing (and iterating), it was easier to have the entire context in the document. I would suggest that we keep it like this until we have a good first full JSON-LD document and then, yes, let's use a URL.

cmaumet:
> the version is a number not a string, but in general we should force that to be a string, since we cannot guarantee appropriate versioning for all software.

Yes, but I think this is true of all strings, not just SPM versioning. e.g. [here](https://github.com/incf-nidash/nidm-specs/blob/8b48784a26039216eac669d1539511e04cea1c46/nidm/nidm-results/spm/spm_results.json#L834). My guess is that maybe compacting does not work when typing is implicit?

satra:
is there a place you could just write the original jsonld file from conversion from rdflib (for one example), as well just the nidmr context file with @version (just for testing purposes)?

cmaumet:
@satra: the context file is saved here: https://github.com/cmaumet/nidm/blob/jsonld_export/nidm/nidm-results/terms/nidmr.json ([raw version](https://raw.githubusercontent.com/cmaumet/nidm/jsonld_export/nidm/nidm-results/terms/nidmr.json)).

cmaumet:
(Also now linked to purl: http://purl.org/nidash/context)

cmaumet:
Also pointed out by @mih
 - [ ] we need to remove multi-typing (cf. https://github.com/incf-nidash/nidmresults-fsl/issues/159 for more context on this issue)

cmaumet:
Hi everyone! I think that this long-lasting pull request is now ready to merge. Thank you to @natachaperez for providing so many fixes!

No changes have been made to the data model, this PR focuses exclusively on updating our set of examples for NIDM-Results.

More specifically, in this PR we:
 - Create a JSON-LD 1.1 context file for NIDM-Results based on the owl file (cf. [`create_nidmr_context.py`](https://github.com/incf-nidash/nidm-specs/pull/479/files?file-filters%5B%5D=.owl&file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-3d8253598057ad57589e0f422dd73a23))
 - Update our scripts to automatically export the NIDM-Results turtle examples into JSON-LD 1.1 (a.k.a. human-readable/nice JSON-LD) (cf. [`create_example_from_templates.py`](https://github.com/incf-nidash/nidm-specs/pull/479/files?file-filters%5B%5D=.owl&file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-1a104299c10d23894780ff8b660e172d))

To ease the conversion, a couple of changes have been made to the turtle templates (used to generate the examples) and the owl file:
 - There is now a "preferred label" attribute for all NIDM-R terms (that are used as keys in the JSON-LD serialisation) (cf. [`nidm-results.owl`](https://github.com/incf-nidash/nidm-specs/pull/479/files?file-filters%5B%5D=.owl&file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-781427a2d36f26d56f2d32045b675d79))
 - To avoid ambiguity when indexing by type in JSON-LD, our turtle examples no longer include the more generic PROV type in addition to the specific NIDM type (cf. e.g. [`ContrastMap.txt`](https://github.com/incf-nidash/nidm-specs/pull/479/files?file-filters%5B%5D=.owl&file-filters%5B%5D=.py&file-filters%5B%5D=.txt#diff-6020c0bd4675ed472916227a92bbe9f7)).
 
Please let us know if you have any comments/feedback. If not, I'll merge at the end of the week. Thanks!

satra:
@cmaumet and @natachaperez - nice work - the one thing i would suggest changing is to first have a PR that generates the context, merge that and then use the purl for that context in the example files. it would be nice to not have really large contexts embedded in the json files.

also to generate the cleanest looking files, you may need to tweak `@context` more, which means updating the converter. but first i would recommend simply playing with updating the context manually in the playground/python console to see the impact on the compact output.

cmaumet:
@satra: Thanks that's a good point! This is now fixed (cf. 78dd6fd, ae60daf). 

I'll update the context purl to redirect to master once this PR is merged.

cmaumet:
Hi everyone! Last call: any additional comment on this PR? Or +1's for merge?

khelm:
I only have a comment on the removal of the prov assignments.  Is the assumption that you'll be able to tell whether the item is an entity, agent, or activity by tracing the term's parents up the tree, so it doesn't need to be explicit?  (This is what I'm assuming in NIDM-E).

cmaumet:
@khelm: yes, that's right.

cmaumet:
I've just updated the context PURL: http://purl.org/nidash/context --now links to--> https://raw.githubusercontent.com/incf-nidash/nidm/master/nidm/nidm-results/terms/nidmr.json.

PR #483, opened 2019-06-13
Add 'deprecated' in labels of deprecated data properties

cmaumet:
This PR was done with @natachaperez. 

Some of our data properties were seen as deprecated in Protege (and possibly by some of our scripts in https://github.com/incf-nidash/nidm-specs/pull/479) when they did not have `owl:deprecated = True`. Digging into this, it looks like their might be bad interactions when a non-deprecated data property has the same `rdfs:label` as a deprecated data property. An example: 

Screen Shot 2019-06-13 at 15 46 54

We have therefore updated the labels of all deprecated data properties to be prefixed by 'deprecated' and this solves the issue: 

Screen Shot 2019-06-13 at 15 56 39


cmaumet:
Once the tests pass, I am going to merge this update that do not impact the data model (changes are only on the labels of deprecated terms).

PR #484, opened 2019-07-29
Add dicom as import

khelm:
This adds a complete list of DICOM tags as datatype properties as well as some terms extracted from the DICOM specification docs.  It also makes a number of other fixes:
1)  In some DICOM tag  description, references are made to pairs of values
in the format #\# or sets of values like "column/row". Unfortunately, the "\" is 
considered by rdflib as an escape character, so if the text is  "-1\-1", rdf lib is 
thinking of this as "\-".  So I changed the "\"'s to "/" which is not an escape character.

The following also use "\", now replaced by "/".  The context is usually "column\row" or "#\#"
`
dicom:dicom_0072003C
dicom:dicom_00181310 
dicom:dicom_00700014
dicom:dicom_00480202
dicom:dicom_00700011
dicom:dicom_00700010
dicom:dicom_00460202
dicom:dicom_00700053
dicom:dicom_00700052
dicom:dicom_00700022
dicom:dicom_00420014
dicom:dicom_00286114
dicom:dicom_00460213
dicom:dicom_60xx0050
dicom:dicom_00221468
dicom:dicom_00221467
dicom:dicom_00720038
dicom:dicom_00480201
dicom:dicom_00480008
dicom:dicom_00181620
dicom:dicom_00289503
`
@cmaumet and @natachaperez:
Also, there are unicode characters in the definitions and, in at least one case, in what is assigned as a key in OwlReader. This still throws an error but it's not from the OWL file, but rather from the OwlReader class in owl_reader.py.  Line 52, where it casts the key as a string - str(key) - is the current error.
Fix by making the cast: unicode(key) in
/home/Karl/Work/packages/miniconda2/envs/nidm-specs/lib/python2.7/site-packages/nidmresults/owl/owl_reader.py
which gets called in owl_to_webpage.py
This fixes the issue.

Added new 
self.owl.graph.bind statements for some of the imported vocabularies (dct, dicom, pato, prov, qibo). 
Already existing in owl_to_webpage.py were (nidm, onli, and sio).


natachaperez:
@khelm ++1 for your job with dicom terms !

PR #488, opened 2020-02-14
Added empty line to retrigger tests

cmaumet:
This a companion PR to https://github.com/incf-nidash/nidmresults/pull/59. To retrigger the tests and check that removing the cast as suggested by @khelm does not cause backward compatibility issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,3 +69,4 @@ We meet every week by videoconference on Mondays at 8-9am PDT / 11am-12pm EDT / 
 
 ## Acknowledgements
 The NIDM working group is a collaboration supported by the [INCF](http://www.incf.org).
+


cmaumet:
@khelm: the tests pass although we do get some missing links -- e.g. to 'Map' in `nidm_results_dev.html`.

khelm:
Thanks @cmaumet - any idea why Map doesn't show up?  This issue arose because there are evidently some non-printing unicode characters in the files that I've been working to eliminate, with only partial success.

cmaumet:
@khelm - I *think* the terms that do not show up are those that are used inside a definition (and were surrounded by quotes so that they would be picked up as terms). 

Do you have an example of where the cast is making the tests fail on nidm-experiment updates? Is https://github.com/incf-nidash/nidm-specs/pull/487 a good example? Or should I be looking somewhere else?

khelm:
Not sure how I missed this...  Yes, #487 is where the issue arises. Thanks!

PR #496, opened 2021-03-10
moving tests to the Github Action workflow

djarecka:
I've moved the tests to Github Action. 

I've started changing the code to work with python3.7, but I've realized that I'd have to change also NIDM Results repo, so perhaps that should be done separately and should start from the NIDM Results repo. Let me know if you would like to do it and need some help.

The other thing that I've noticed is that running the tests (on my OSX with python 2.7) changes some files from the repo. Perhaps some files need updates.


cmaumet:
Hi @djarecka! My apologies for the slow response. Thanks a lot for working on this! 

> I've started changing the code to work with python3.7, but I've realized that I'd have to change also NIDM Results repo, so perhaps that should be done separately and should start from the NIDM Results repo. Let me know if you would like to do it and need some help.

I've just checked and the nidmresults repository should already be Python 2 and Python 3 compatible (see [here](https://github.com/incf-nidash/nidmresults/blob/master/.travis.yml#L7-L9)).

> The other thing that I've noticed is that running the tests (on my OSX with python 2.7) changes some files from the repo. Perhaps some files need updates.

This has happened to both Karl and myself in the past and is usually due to a version of our dependencies being different locally compared to what is used by Travis for online testing (in particular the version of the markdown package). The current build in master is passing which means that the tests did not find any difference between what has been pushed and what is recomputed.

Let me know if I can provide more info. Thanks!

djarecka:
@cmaumet - also have some delays, sorry for this. I think I mistaken your repo with `nidm-specs/nidm/nidm-results/scripts/`.

PR #501, opened 2021-07-02
Added Experiment Expanded to Readme

wolfborg:
Changed the README.md file to also include the link to the new NIDM-Experiment Expanded


khelm:
Thanks Derek!  I will merge and then create the webpages.  I also noticed that there are a number of errors that pop up on http://nidm.nidash.org/specs/nidm-experiment_dev.html (click on the orange and red boxes at the upper right of the page) I will see if they originate in the original files or later in the conversion and let you know.

